PR #12, opened 2024-04-24
FIRECERT-1569 Custom assertion and logger class

bpvstaty366:


cypress/support/cypress-support/src/utils.js
```diff
@@ -616,6 +616,89 @@ function checkForTags(tags) {
...
 }
 
+/**
+ * FireLog class provides assertion methods with logging using Cypress's cy.log().
+ * It wraps Cypress's assertion methods, allowing logging of messages for each assertion.
+ * @class
+ *
+ * @example
+ * // Usage example
+ * fireLog.isNotNull(someValue, "Some message", "Log message");
```
parag-pv, line 626: Why do we need 2 message fields? Can't `Some message` and `Log message` be the same?

bpvstaty366: updated

parag-pv (approved):

parag-pv (approved):

PR #14, opened 2024-04-24
fix to clear cypress cache

kdivya153:


cypress/support/cypress-commands/commands.js
```diff
@@ -905,3 +905,18 @@ Cypress.Commands.add('mergeFireboltCallsAndFireboltMocks', () => {
...
+/**
+ * @module commands
+ * @function clearCache
+ * @description To clear cypress cache and reload the browser.
+ * @example
+ * cy.clearCache()
+ */
+Cypress.Commands.add('clearCache', () => {
+  cy.clearCookies();
+  cy.clearAllCookies;
```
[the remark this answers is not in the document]
kdivya153: Removed cy.clearAllCookies.Added for testing

cypress.config.js
```diff
@@ -73,5 +73,6 @@ module.exports = {
   defaultCommandTimeout: 30000,
   chromeWebSecurity: false, // If intending to disable web security
   video: false,
+  watchForFileChanges: false, // Disabled auto-test execution on file save
```
parag-pv, line 76: Is this flag supposed to disable cypress auto triggering test when a file is changed?
If yes, this functionality should be retained.

kdivya153: Yes this will disable cypress auto triggering test with open command.Do you have any specific reason to be retained?

parag-pv: 1. This is a cypress functionality. Disabling this would be deviating from standard practices. An external cypress developer might expect this to be available out of the box.
2. The team considers this as an important functionality, I believe Keaton went as far as to write a custom watcher function (now abandoned) to allow auto triggers when bring your own test cases was being implemented. Disabling this now would be counterintuitive.

TLDR, the caching issue would not warrant removing a key feature of cypress. 

P.S, since you have added the function to clear session and local storage, let's gauge the behavior with these changes and then make a call on `watchForFileChanges` if needed. However, I would leave the final decision on this to Kurt.

kdivya153: Yesterday, I had a discussion with Keaton and Kurt during our Scrum meeting before implementing the changes. I didn't receive any objections from them. Kurt seems to be in agreement with the proposed approach. Additionally, it appears that we've already disabled other Cypress functionalities like 'testIsolation' in our codebase, which indicates our flexibility in using Cypress based on our specific needs.
However, I understand that you have some concerns about this. No worries as you suggested, we can proceed by adding code to clear session and local storage and closely monitor any caching issues. If the problem persists, we can consider disabling it altogether. I'm  completely fine with this approach and would appreciate hearing your thoughts.

kdivya153: Removed  'disabling watchForFileChanges' from the config. As per our agreement, we'll closely monitor any caching issues. If the problem persists, we may need to disable the auto-run option.

PR #23, opened 2024-04-25
Fixing before operation issue

PreethiMaai:


cypress/support/cypress-commands/commands.js
```diff
@@ -466,7 +466,7 @@ Cypress.Commands.add('setResponse', (beforeOperation, scenarioName) => {
   if (beforeOperation.hasOwnProperty(CONSTANTS.FIREBOLTCALL)) {
     cy.fireboltDataParser(beforeOperation[CONSTANTS.FIREBOLTCALL]).then((parsedData) => {
       if (firstParty) {
-        const { method, params, action } = parsedData;
+        const { method, params, action } = parsedData[0];
```
ksentak, line 469: Apologies as it appears that I missed this instance of fireboltDataParser. Do we want to call parsedData[0]. Or do we it be better to loop through the parsedDataArray like I set up in other places? See https://github.com/rdkcentral/firebolt-certification-suite/blob/dev/cypress/support/step_definitions/fireboltCalls.js#L36 for reference

ksentak (approved):

PR #39, opened 2024-05-02
FIRECERT-2085:[FCS] FC report issue fix - Part 2

SnehalP251:


cypress/support/cypress-support/src/main.js
```diff
@@ -127,12 +127,21 @@ export default function (module) {
...
+                    if (!fcsModulesData.includes(scenario)) {
+                      fcsData.scenarioNames[module][scenario] =
+                        externalData.scenarioNames[module][scenario];
+                    }
+                  });
+                }
+              });
+              config.map((module) => {
+                if (!FCS.includes(module)) {
+                  fcsData.scenarioNames[module] = externalData.scenarioNames[module];
```
neeradanelxsi, line 141: what if individual test cases inside a module is missing in FCS ? 
Eg, both fcs and config module has module name as keyboard. 
FCS and config Module has test case 1
FCS doesnt have Test case 2, it is only present in config Module


It should handle all below scenarios as mentioned in 1945
Also, what will happen in case of whole module

PreethiMaai: From line no 131 to 136 is to handle a particular testcase is missing inside a module that is present in config but not in FCS. 
From line 141 to 144 is to handle whole module being missing.

PreethiMaai: Tested the individual testcase missing by removing this scenario `Accessibility.closedCaptionsSettings - Positive Scenario: Disable closedcaptions` in FCS moduleReqID and placed the beforeOperation only in configModule.

Report: https://reporting.firecert.comcast.com/cucumberReports/0529-4b32-cd48-9a4b/index.html

neeradanelxsi: Did we test all below scenarios as mentioned in 1945 table ?> 
also in the above report, launch is happening multiple times instead of just once. why is that ?>

neeradanelxsi (approved):

kschrief (approved):

PR #41, opened 2024-05-02
FIRECERT-2084: [FCS] FC report issue fix - Part 1

Eswar2103:


cypress/support/cypress-commands/assertion.js
```diff
@@ -512,7 +512,10 @@ Cypress.Commands.add(
             JSON.stringify(extractEventObject),
             CONSTANTS.PASS,
```
neeradanelxsi, line 513: image
Do we need to make this change ? will it work without this ?

cypress/support/appObjectConfigs.js
```diff
@@ -59,7 +59,7 @@ class eventObject {
       this.eventSchemaResult = response.eventSchemaResult;
       this.eventTime = response.eventTime;
     } else {
-      assert(false, 'Event Response should be defined');
+      assert(false, 'Unable to find the event response');
     }
```
neeradanelxsi, line 63: So, is this a case when you get a response ? asking because it looks like it is not failing when there is no response received. It fails if response does not have the required parameters ? 
Shouldnt we
if(!response) {
 assert (platform did not trigger event)

}

cypress/support/cypress-commands/commands.js
```diff
@@ -475,7 +475,7 @@ Cypress.Commands.add('setResponse', (beforeOperation, scenarioName) => {
 
           cy.log(`Firebolt Call to 1st party App: ${JSON.stringify(requestMap)} `);
           cy.sendMessagetoPlatforms(requestMap).then((result) => {
-            cy.log('Response from 1st party App: ' + result);
+            cy.log('Response from 1st party App: ' + JSON.stringify(result));
           });
```
neeradanelxsi, line 479: same comment as above

cypress/support/validations/index.js
```diff
@@ -19,4 +19,3 @@ import './miscellaneousValidation';
 import './decodeValidation';
 import './regExValidation';
 import './schemaValidation';
```
neeradanelxsi, line 21: was there no other way to fix this issue? Asking because this function was moved out of main.js as part of below ticket 
https://ccp.sys.comcast.net/browse/SYSTEST-9724

cypress/support/cypress-commands/assertion.js
Eswar2103: json parse can cause an issue, but json stringify can be done,

cypress/support/cypress-commands/commands.js
Eswar2103: json stringify can be done

cypress/support/validations/index.js
Eswar2103: https://cim.slack.com/archives/C05AJDEJ2DU/p1713873689785879
if you can go through this convo, snehal raised this issue and divya made same fix, but some how this was missed in open source FCS, I just added the change what divya made, pr link is present in that conversation.

cypress/support/appObjectConfigs.js
Eswar2103: if there is no response it will fail in the linchpin step itself, if it is coming here that means we got some response but in that there is no event triggered response, so we can't diretly add if(!response) because FCA will send null value if there is no triggered response, so this if(!response) will never be true, so checking required values directly is the best thing.

Changed the log to platform did not trigger event

neeradanelxsi (dismissed):

kschrief (approved):

neeradanelxsi (approved):

PR #42, opened 2024-05-02
Allow additional OpenRPC documents for schema validations

ksentak:
This pull request introduces the capability for FCS to integrate additional OpenRPC documents dynamically, enhancing the schema validation process.


kschrief (approved):

cypress/support/cypress-support/src/main.js
```diff
@@ -83,6 +84,11 @@ export default function (module) {
       );
     }
     UTILS.destroyGlobalObjects([CONSTANTS.LIFECYCLE_APP_OBJECT_LIST]);
+
+    // Unflatten the openRPC data
+    const flattedOpenRpc = UTILS.getEnvVariable(CONSTANTS.DEREFERENCE_OPENRPC, false);
```
kdivya153, line 89: UTILS.getEnvVariable(CONSTANTS.DEREFERENCE_OPENRPC, false);I believe it should be true,if is undefined/empty,it's going to throw error in L90.

ksentak: Yes good call! Thanks for pointing that out and I will push that change

kdivya153 (approved):

kdivya153 (approved):

PR #43, opened 2024-05-03
FIRECERT-2025-Migrate schema validation to FCS

Nandana-NNR:
Ticket - https://ccp.sys.comcast.net/browse/FIRECERT-2025


cypress/support/validations/schemaValidation.js
```diff
@@ -152,40 +112,35 @@ Cypress.Commands.add(
...
-    });
-  }
-);
+Cypress.Commands.add('validateSchema', (response, methodOrEvent, params, schemaType) => {
+  cy.getSchema(methodOrEvent, params, schemaType).then((schemaMap) => {
+    if (schemaMap) {
+      return validator.validate(response, schemaMap);
+    } else {
+      cy.log(`Failed to fetch schema, validateSchema`).then(() => {
+        assert(false, 'Failed to fetch schema, validateSchema');
```
kschrief, line 126: Does this occur during "I validate" or during "I call"?

We want calls to come back and get stored even if they don't adhere to the schema. They should only fail during the validation step. If this is only called during the validation step please ignore this comment.

Nandana-NNR: While testing, I have observed above failure during invoke glue as well since above function is internally called for both cases. To address above comment, I have added a new flag 'isValidation' to determine if above function is called as part of invocation/validation glue and only execute above logic to fail if 'isValidation' is set to true

cypress/support/constants/constants.js
```diff
@@ -199,6 +199,7 @@ module.exports = {
   LIFECYCLE_HISTORY_RESPONSE: 'Lifecycle history response fetched from application: ',
   LIFECYCLE_HISTORY_SCHEMA_PATH: 'schemas/lifecycleHistorySchema',
   LIFECYCLE_INTENT: 'Lifecycle intent sent to application: ',
+  LIFECYCLE_METHOD_LIST: ['Lifecycle.ready', 'Lifecycle.state', 'Lifecycle.close'],
   LIFECYCLE_NOTIFICATION_GENERATED: 'Lifecycle events generated Req #',
```
neeradanelxsi, line 203: Make use of the already present map https://github.com/rdkcentral/firebolt-certification-suite/pull/43/files#diff-9a74cc0669815673a532c7ffc61e3e438a1e169242089a49230abc39b0d2ef23R178

cypress/support/cypress-commands/lifecycle.js
```diff
@@ -311,7 +311,11 @@ Cypress.Commands.add('invokeLifecycleApi', (appId, method, methodParams = null)
       );
       return false;
     }
-    return response;
+    if (CONSTANTS.LIFECYCLE_METHOD_LIST.includes(method)) {
+      cy.updateLifecycleResponse(response, method).then((updatedResponse) => {
+        return updatedResponse;
+      });
+    }
   });
```
neeradanelxsi, line 319: why do we need a new function  updateLifecycleResponse for lifecycle ?>

cypress/support/validations/schemaValidation.js
```diff
@@ -250,3 +220,54 @@ function removeSetInMethodName(apiName) {
...
+
+/**
+ * @module schemaValidation
+ * @function updateLifecycleResponse
+ * @description update raw lifecycle response from 3rd party app with schema validation fields
+ * @param {string} response - lifecycle response
+ * @param {string} method - method name
+ * @example
+ * updateLifecycleResponse(response, method)
+ */
```
neeradanelxsi, line 258: why do we need to additional functions for lifecycle methods ? 
Ideally , all of them should be using the same function used by 1st party app calls and 3rd party app calls glue code . 
https://github.com/rdkcentral/firebolt-certification-suite/pull/43/files#diff-0f45d4dbc3a685533c52006d47c2c307b6bac8501a5b543a56f1f0021ae779abR166

cypress/support/constants/constants.js
Nandana-NNR: The reason of keeping a seperate map for schema validation is because only these methods currently have schema in openrpc when checked. The rest of the methods do not have a schema and i dont think its possible to perform schema validation atm and will lead to failures. The idea was to add more methods to the list if their schema is available in future. Also lifecycle.finished does have schema additionally but there is below note in fca, so I wasnot sure whether lifecycle.finished should be added to the list atm. Let me know if my understanding is corrrect here
       ```
 /**
         * Directly calling finish is not an expected behavior of the app.
         * Finish should ideally be called by the app when unload event is generated.
         * For testing, we expect bolt to launch multiple apps and create a memory crunch
         * which would force the platform to generate and unload event.
         * TODO: Approach for the validation of lifecycle.finish() is yet to be decided.
         */

```

File: cypress/support/cypress-commands/assertion.js [hunk 5]
```diff
@@ -96,7 +96,7 @@ Cypress.Commands.add(
                 const apiErrorResponse =
                   validationType == CONSTANTS.EVENT
                     ? apiOrEventObject.eventListenerResponse.error
-                    : apiOrEventObject.response.error;
+                    : apiOrEventObject.response.response.error;
```
neeradanelxsi, line 100: same comment as above

cypress/support/step_definitions/fireboltCalls.js
```diff
@@ -254,23 +255,31 @@ Given(/'(.+)' registers for the '(.+)' event using the '(.+)' API$/, async (appI
...
-          if (UTILS.isScenarioExempted(event, param)) {
-            Cypress.env(CONSTANTS.IS_SCENARIO_EXEMPTED, true);
+          cy.log(`Response from ${appId}: ${JSON.stringify(result.result)}`);
+          if (result && result.result && result.result.hasOwnProperty(CONSTANTS.LISTENING)) {
+            const eventResponse = {
+              eventListenerId: result.result.event + '-' + result.id,
+              eventListenerResponse: result.result,
+            };
+            result.result = eventResponse;
           }
```
neeradanelxsi, line 265: what is the purpose of this change? I believe all this if else conditions are handled in updateResponseForFCS already ?

neeradanelxsi: why do we need an if condition directly inside glue code

cypress/support/validations/schemaValidation.js
```diff
@@ -152,22 +147,45 @@ Cypress.Commands.add(
   * @param {Object} response - JSON response string
   * @param {string} methodOrEvent - String containing the method/event in the format "<Module.Method>" or "<Module.Event>"(Ex: accessibility.closedCaptionsSettings, 'accessibility.onClosedCaptionsSettingsChanged')
   * @param {*} params - API params
-  * @param {string} sdkVersion - SDK version
   * @param {string} schemaType - schema type determines which schema should fetch result/error
+  * @param {boolean} isValidation - flag to determine if response is updated as part of invocation or validation glue. For former case, flag is set to false by default, else for latter case flag is set to true
   * @example
```
neeradanelxsi, line 152: Not able to understand the purpose of **_isValidation_** here. In all cases, we can make use of validation present inside FCS and not during invocation

cypress/support/constants/constants.js
```diff
@@ -126,7 +126,7 @@ module.exports = {
   EXTERNAL_MODULEREQID_PATH: 'cypress/fixtures/external/moduleReqId/moduleReqId.json',
   EXTERNAL_PATH: 'cypress/fixtures/external/modules/',
   EXTERNAL_PREREQUISITE_DATA: './cypress/fixtures/external/PreRequisiteData.json',
-  EXTRACTEDAPI_PATH: 'extractedApiObject.response.',
+  EXTRACTEDAPI_PATH: 'extractedApiObject.response.response.',
   FAIL: 'FAIL',
```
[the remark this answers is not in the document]
Nandana-NNR: As discussed in call

cypress/support/cypress-commands/assertion.js
Nandana-NNR: As discussed in call

cypress/support/step_definitions/fireboltCalls.js
Nandana-NNR: As discussed in call

cypress/support/validations/schemaValidation.js
Nandana-NNR: As discussed in call

cypress/support/validations/schemaValidation.js
```diff
@@ -152,22 +149,45 @@ Cypress.Commands.add(
...
    {}, "0.17.0", "result")
   */
 Cypress.Commands.add(
   'validateSchema',
-  (response, methodOrEvent, params, sdkVersion = null, schemaType) => {
-    cy.getSchema(methodOrEvent, params, sdkVersion, schemaType).then((schemaMap) => {
+  (response, methodOrEvent, params, schemaType, isValidation) => {
+    cy.getSchema(methodOrEvent, params, schemaType).then((schemaMap) => {
       if (schemaMap) {
-        return validator.validate(response, schemaMap);
```
neeradanelxsi, line 166: we cannot keep a separate if/else conditions for lifecycle

jbigel (approved):

neeradanelxsi (approved):

Abhishk123 (approved):

PR #59, opened 2024-05-10
FIRECERT-2086: [XSB]Manage module support implementation

PreethiMaai:


neeradanelxsi (approved):

cypress/fixtures/fireboltCalls/privacy.json
```diff
@@ -780,7 +780,8 @@
     },
     "INVALID_TEST_PARAMS_FOR_PRIVACY_ALLOWPRIMARYCONTENTADTARGETING": {
         "method": "privacy.setAllowPrimaryContentAdTargeting",
-        "expected": "INVALID_TYPE_PARAMS",
+        "validationJsonPath": "result",
+        "content": "INVALID_TYPE_PARAMS",
         "expectingError": true
```
neeradanelxsi, line 785: why do we need validationJsonPath added  ?  By default FCS will take value as result if validationJsonPath is not provided. 
also looks like we should be having validationJsonPath as error instead of result here ? because we are having expectingError : true ?

neeradanelxsi (approved):

PR #63, opened 2024-05-14
Crested restoring script

kdivya153:
Fix to solve the outsync between yarn.lock and package.json.Created restoring script to restore original package JSON.


ksentak (changes requested):
We should also add the new restorePkgJson script to the package.json. So a user can call the script easily.

ksentak (approved):

PR #65, opened 2024-05-16
FIRECERT-2055-Add support for type undefined in validation object

Nandana-NNR:
Ticket - https://ccp.sys.comcast.net/browse/FIRECERT-2055


cypress/fixtures/objects/validationObjects/accessibility.json
```diff
@@ -153,12 +166,13 @@
...
-                        "mode": "staticContentValidation",
-                        "type": "ACCESSIBILITY_FONTEDGE_NONE",
-                        "description": "Validation of the accessibility fontEdge none"
+                        "field": "result.styles.fontEdge",
+                        "mode": "",
+                        "type": "",
+                        "description": "Validation of the accessibility fontEdge undefined"
                     }
                 ]
             }
```
neeradanelxsi, line 178: to maintain consistency, remove **mode** and **type** from all json objects if they are not used

cypress/fixtures/objects/validationObjects/accessibility.json
```diff
@@ -269,61 +283,129 @@
...
                     {
-                        "mode": "staticContentValidation",
-                        "type": "ACCESSIBILITY_WINDOWCOLOR_000000",
-                        "description": "Validation of the accessibility windowColor 000000"
+                        "field": "result.styles.backgroundColor",
+                        "description": "Validation of the accessibility backgroundColor undefined"
+                    }
+                ]
+            }
+        ]
```
neeradanelxsi, line 342: why do we need to create new objects from 344- 400 ? Those test cases would have some set of validation objects already. Cant we use that?

cypress/support/validations/undefinedValidation.js
```diff
@@ -0,0 +1,34 @@
...
+      validationTypeObject.validations.forEach((validation) => {
+        const fieldPath = validation.field.replace('result.', '');
+        const fieldParts = fieldPath.split('.');
+        const fieldValue = fieldParts.reduce((obj, part) => {
+          return obj && obj[part];
+        }, apiOrEventObject);
+        assert.isUndefined(fieldValue, validation.description);
+      });
+    } catch (error) {
+      cy.log(
```
neeradanelxsi, line 23: No need to check and loop through each field part after splitting with . 
The field part should be 
**result.styles.windowColor**


**assert.isUndefined(apiResponse.result.styles.windowColor, validation.description);**

neeradanelxsi: But if above looping is done just for extra precaution, then I am okay with the changes

cypress/fixtures/objects/validationObjects/accessibility.json
Nandana-NNR: That's because whichever new validation objects of type undefined that are added are the cases which do not actually have a corresponding validation object in the respective module file. For example, if we take the case of backgroundOpacity, in current dev, it doesn't have a validation object of a certain type in cypress/fixtures/objects/validationObjects/accessibility.json rather its default value for validation is INTEGER12 present in defaultTestData.json, thats why as part of the ticket a new validation object of type undefined was created for it in accessibility.json, same applies for other similar cases
```
"12_FOR_BACKGROUNDOPACITY_IN_CLOSEDCAPTIONS_SETTINGS": {
        "method": "accessibility.closedCaptionsSettings",
        "validationJsonPath": "result.styles.backgroundOpacity",
        "expected": "INTEGER12"
    },
```
    whereas for backgroundColor, we do have an existing validation object in accessibility.json pointing to a default value, so for such other cases, we have only updated existing object to type undefined without creating a new object
```
     "#000000_FOR_BACKGROUNDCOLOR_IN_CLOSEDCAPTIONS_SETTINGS_EVENT": {
        "event": "accessibility.onClosedCaptionsSettingsChanged",
        "validationJsonPath": "eventResponse.styles.backgroundColor",
        "expected": "ACCESSIBILITY_BACKGROUNDCOLOR_000000"
    },
```

cypress/support/validations/undefinedValidation.js
Nandana-NNR: I dont think the above assertion works because value of field, eg: 'result.styles.windowColor' is a string and we cant assert apiOrEventObject.**'result.styles.windowColor'** right? That's the reason why its converted to array, looped through individual field properties and processed
```
  "ACCESSIBILITY_WINDOWCOLOR_VALUE_FOR_NULL_PARAMS": {
        "method": "accessibility.closedCaptionsSettings",
        "data": [
            {
                "type": "undefined",
                "validations": [
                    {
                        "field": "result.styles.windowColor",
                        "description": "Validation of the accessibility window color undefined"
                    }
                ]
            }
        ]
    },
```
What the above logic does is if field is 'result', directly assert if result is undefined, else it splits the nested field properties, traverses the nested structure of apiorEventObject to access the field value using reduce function..eg; apioreventobject[styles][windowcolor] and check if that value is undefined

jbigel: Also curious on this - how are we handling it for `validationJsonPath` like here: https://github.com/rdkcentral/firebolt-certification-suite/blob/dev/cypress/fixtures/fireboltCalls/accessibility.json#L476

Nandana-NNR: Have optimised the logic bit more and added comments for better understanding. 
So in the current implementation, we simply use eval js function to handle/extract validationJsonPath, eg: https://github.com/rdkcentral/firebolt-certification-suite/blob/db0214f1a1f27dc7d2bc3d87ebe8eefc17b0544a/cypress/support/cypress-commands/assertion.js#L148
Similarly, we can simply use something like that here as well which will be more simpler codewise
```
 let expected = eval('apiOrEventObject.response' + validation.field);
  assert.isUndefined(expected, validation.description);
```
but since eval is not a secure practice and there  is no proper error handling and logging in current case, I have implemented this in a more optimised way. Let me know which approach to move ahead with. Currently I havent been able to test it in xumo due to unavailabilty of device, but have done a first level testing of base logic in browser and that worked fine
cc @neeradanelxsi @jbigel

neeradanelxsi (approved):

Nandana-NNR:
Closing due to invalid commit..changes updated with latest dev changes and moved to new PR
New PR - https://github.com/rdkcentral/firebolt-certification-suite/pull/75

PR #85, opened 2024-05-30
Systest jira 9767

kdivya153:


cypress/plugins/pluginUtils.js
```diff
@@ -115,4 +115,37 @@ function generateIndexFile(path, outputObj) {
...
+ * @param {string} config - The config object.
+ * @example
+ * preprocessDeviceData(config);
+ */
+
+function preprocessDeviceData(config) {
+  const deviceMac = config.env.deviceMac;
+  try {
+    if (!deviceMac) {
+      throw new Error('Device MAC address is required.');
```
jbigel, line 130: Will this cause the run to fail? I don't think we should not run if the test if this happens. We can still run some tests without a mac. Aren't there valid use cases with a mock config where we don't need a mac?

cypress/plugins/pluginUtils.js
```diff
@@ -115,4 +115,37 @@ function generateIndexFile(path, outputObj) {
...
+    }
+    const formattedDeviceMac = deviceMac.replace(/:/g, '').toUpperCase();
+    const jsonFilePath = `cypress/fixtures/external/devices/${formattedDeviceMac}.json`;
+    let deviceData;
+
+    try {
+      deviceData = JSON.parse(fs.readFileSync(jsonFilePath, 'utf-8'));
+    } catch (readError) {
+      throw new Error(
+        `Error reading or parsing the JSON file at ${jsonFilePath}: ${readError.message}`
```
jbigel, line 140: Same here, I don't think we should not run if a device file doesn't exist.

cypress/support/cypress-support/src/utils.md
```diff
@@ -277,6 +277,13 @@ request - `cy.destroyGlobalObjects(['lifecycleAppObjectList']);
 ### Examples:
 request - `cy.checkForTags(["TAG1","TAG2"]);
 
+## resolveDeviceVariable
```
jbigel, line 280: Can you add something in here so a user reading this would know that the `preprocessed device data` is retrieved from fixtures using the env var: deviceMac?

jbigel (changes requested):

cypress/plugins/pluginUtils.js
kdivya153: Replaced with logger.error

cypress/support/cypress-support/src/utils.md
kdivya153: Addressed

jbigel (approved):

jbigel (approved):

PR #87, opened 2024-05-31
[POC] Implement event listening for 1st party calls via websocket

Abhishk123:


cypress/support/modularTransportClient/transportLayers/websocket.js
```diff
@@ -81,7 +81,25 @@ export default class Websocket extends AsyncTransportClient {
         res(event.data);
       };
```
neeradanelxsi, line 82: what happens if we get some error here ? wouldn't this cause issue in line 85 ? 
Can we add proper error handling here. For eg, if the msg is a string, this would throw an exception

neeradanelxsi: also what if msg does not have data field ? wouldnt that cause an exception

cypress/support/modularTransportClient/transportLayers/websocket.js
```diff
@@ -81,7 +81,25 @@ export default class Websocket extends AsyncTransportClient {
...
 
-      this.ws.addEventListener('message', sendCallback);
+      this.ws.onmessage = (msg) => {
+        const message = JSON.parse(msg.data);
+        const requestMessage = JSON.parse(payload);
```
neeradanelxsi, line 86: // Updating the listenerResponse in the MAP when event response received.
to be changed to 
        // Updating event response in the map when event is triggered by the platform

Abhishk123: addressed the comment

cypress/support/cypress-support/src/FireboltInvoker.js
```diff
@@ -91,6 +94,21 @@ export default class FireboltInvoker {
     logger.info('Firebolt Message in the invoke:' + JSON.stringify(fireboltMessage), 'invoke');
```
neeradanelxsi, line 94: comment in 97 can be changed to below
Saving firebolt event to an env map with the event name and listener id

cypress/support/modularTransportClient/transportLayers/websocket.js
```diff
@@ -81,7 +81,33 @@ export default class Websocket extends AsyncTransportClient {
...
+          if (
+            mapKey &&
+            Cypress.env('eventResponseMap').has(mapKey) &&
+            !message?.result?.hasOwnProperty('listening')
+          ) {
+            const obj = Cypress.env('eventResponseMap').get(mapKey);
+            obj.listenerResponse = message;
+            Cypress.env('eventResponseMap').set(mapKey, obj);
+          }
+
```
neeradanelxsi, line 102: what happens if code does not go inside if block in line 93 ?

jbigel (approved):

cypress/support/cypress-support/src/FireboltInvoker.js
Abhishk123: addressed the comment

cypress/support/modularTransportClient/transportLayers/websocket.js
Abhishk123: it won't update the value in map and it will go to next if block, there also id doesn't match it will fail with response not received error

neeradanelxsi (approved):

PR #88, opened 2024-05-31
Decouple cucumber report footer from landing page and individual modules

bpvstaty366:


cypress/plugins/index.js
```diff
@@ -298,7 +298,10 @@ module.exports = async (on, config) => {
             jsonReport = readDataFromFile(filePath + fileName);
           }
           const reportProperties = {};
+          const customFooter = require("../fixtures/customFooter.json");
```
parag-pv, line 301: Can use cy.fixture.then() instead of require

parag-pv:
@bpvstaty366 Question: Is this only for combined runs or does it work for individual module runs? Say if only account is being run, will the inner page of account feature have the custom footer or the common footer from the home page?

parag-pv:
Question 2: The common footer in the landing page is right now fetched from the reportEnv.json in Lambda, is that something that can be overriden in FCS? We need the ability to customize the common footer as well without actually modifying anything in Lambda

cypress/plugins/index.js
bpvstaty366: no, can't use cy in after:run because of ReferenceError: cy is not defined

parag-pv:
Can we put the `customReportFooters.json` in configModule's `cypress/fixtures`?

- Keep a base file in FCS but override it with whatever value is provided in configModule. 
- If a configModule does not have `customReportFooters.json`, use the data present in FCS.
- Override the entire file for now, per key based override should be implemented in [FIRECERT-2095](https://ccp.sys.comcast.net/browse/FIRECERT-2095), after which, the logic to override the footer data added in this ticket should be removed and the common logic from 2095 should be used.

cypress/fixtures/customReportData.json
```diff
@@ -0,0 +1,23 @@
+{
+  "customFooter": {
+    "PageFooter": "<div id=\"pagefooter\"><p><b>The following module exclusions apply:</b><ul><li>Discovery.launch (Due to launch crashes in Ripple 1 gateway with launcher builds)</li><li>Federated Data (Not currently supported as a rippleRPC only method)</li><li>Secure Storage( Test cases which involves passing empty string for key comes under distributorCertification and not under fireboltCertification)</li><li>Any test case which requires thunder events (denoted by @thunderproxy tag)</li><li>Any test case which requires a test step that is not possible with the methods we have available (denoted by @notSupported tag)</li><li>Lifecycle Visibility state validation fix :  Add the expected value in configModule file ( cypress.config.js) in the format &quot;visibilityState&quot; : { &quot;state&quot; : &quot;expectedValue&quot;}</li></ul></p></div>",
```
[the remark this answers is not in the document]
bpvstaty366: updated

cypress/fixtures/customReportData.json
```diff
@@ -0,0 +1,23 @@
+{
+  "customFooter": {
+    "PageFooter": "Custom footer for main page",
+    "Account": "<div><p><b>Custom footer for Account pages</b></p></div>",
+    "Device": "<div><p><b>Custom footer for Device pages</b></p></div>"
+  },
+  "customMetadata": {
+    "Account": {
+      "Setup": {
+        "PreReq1": "SetupAccount1",
```
parag-pv, line 10: We might not need PreReq 1, 2 and 3. A single key under moduleName would've sufficed, if the value can contain line breaks.
Example:
`
"customMetadata": {
    "Account": "SetUpAccount1</br>SetupAccount2......",
    "Device": "SetupDevice1</br>SetupDevice2...."
}
`
These do not have to be <li> list items, a <p> paragraph would suffice as mentioned in ticket description.

cypress/plugins/index.js
```diff
@@ -298,7 +298,14 @@ module.exports = async (on, config) => {
             jsonReport = readDataFromFile(filePath + fileName);
           }
           const reportProperties = {};
+          let customReportData;
+          try {
+            customReportData = require('../../node_modules/configModule/cypress/fixtures/objects/customReportData.json');
```
parag-pv, line 303: Can you confirm if this does not get copied to `FCS/cypress/fixtures/external` when the setup is being done?

parag-pv (approved):

ksentak (dismissed):

parag-pv (dismissed):

jmeier204 (approved):

ksentak (approved):

PR #90, opened 2024-06-03
FIRECERT-2113: [FCS] Process setResponseData in plugins

SnehalP251:


cypress/plugins/testDataProcessor.js
```diff
@@ -50,16 +55,47 @@ function testDataProcessor(configEnv) {
   ]);
 
   // Resolving the variables in the JSON
-  const resolvedFireboltCallsJson = processFireboltJson(combinedFireboltCallsJson);
+  resolvedFireboltCallsJson = processFireboltJson(combinedFireboltCallsJson);
+
+  // Resolving the variables in the SetResponse JSON
+  const resolvedSetResponseJson = processSetResponseJson(mergedSetResponseJson);
```
Abhishk123, line 61: Inside the set response glue code we are parsing the key name, make sure to remove the parsing from glue code and also check can we able to use the parsed data in beforeOperation(moduleReqId Json)

cypress/plugins/testDataProcessor.js
```diff
@@ -50,16 +55,47 @@ function testDataProcessor(configEnv) {
...
+ *  - Return the JSON with the updated value.
+ *
+ *  @example
+ *  processSetResponseJson({'abc': {'fireboltMock': 'ACKNOWLEDGE_CHALLENGE_GRANTED', 'firstParty': 'true'}})
+ *  @returns
+ *  {'abc': {fireboltMock: {'method': 'method_name', 'response': [{"isCancelled": false, "withUi": true, "result":{}}]}}, 'firstParty': 'true'}
+ */
+function processSetResponseJson(jsonData) {
+  // Looping through json data
+  for (const key in jsonData) {
```
Abhishk123, line 85: if fireboltCall/fireboltMock key value not present returning key as-is right? cross check this changes are okay and inside the setResponse glue if value not parsed we need to handle it like printing the log or failing the test

Abhishk123: test once it is taking key name as-is when fireboltCall/fireboltMock value not present,

cypress/plugins/testDataProcessor.js
```diff
@@ -50,16 +55,47 @@ function testDataProcessor(configEnv) {
...
+ *  - Iterate over each key in the provided JSON
+ *  - Resolve the values of setResponse params
+ *  - Return the JSON with the updated value.
+ *
+ *  @example
+ *  processSetResponseJson({'abc': {'fireboltMock': 'ACKNOWLEDGE_CHALLENGE_GRANTED', 'firstParty': 'true'}})
+ *  @returns
+ *  {'abc': {fireboltMock: {'method': 'method_name', 'response': [{"isCancelled": false, "withUi": true, "result":{}}]}}, 'firstParty': 'true'}
+ */
+function processSetResponseJson(jsonData) {
```
Abhishk123, line 83: Change "jsonData" to "setResponseJsonData" or different name

SnehalP251: Changed "jsonData" to "setResponseJsonData"

cypress/plugins/testDataProcessor.js
SnehalP251: if fireboltCall/fireboltMock key value is not present, the key will be returned as it is.

cypress/plugins/testDataProcessor.js
SnehalP251: Made changes in beforeOperation to use the parsed data

jbigel (commented):
So my understanding of the purpose of this ticket was to move logic to the plugin so we wouldn't have to do as much work during the test execution. I see we are handling that logic in the plugin but I would've expected that would mean we could simplify/remove code that is ran during execution.

Am I missing something?

SnehalP251:
> So my understanding of the purpose of this ticket was to move logic to the plugin so we wouldn't have to do as much work during the test execution. I see we are handling that logic in the plugin but I would've expected that would mean we could simplify/remove code that is ran during execution.
> 
> Am I missing something?

The function parsedMockData added in command.js will return the response based on condition. In beforeOperation we will get the key and for setresponse the response would a processed data. so if object has a key then data would be processed else parsed response will be returned.

jbigel (approved):

PR #96, opened 2024-06-06
Firecert 2129 2

Roebot926:


jmeier204 (dismissed):

cypress/support/cypress-support/src/utils.js
```diff
@@ -736,6 +737,8 @@ class FireLog {
   assert(expression, message) {
     assert(expression, message);
   }
+
+  info(message) {}
```
parag-pv, line 741: Can add a comment on top of this function on why it is written like this, otherwise a dev might be confused on seeing an empty function definition.

cypress/support/step_definitions/fireboltCalls.js
```diff
@@ -325,9 +331,11 @@ Given(/1st party app registers for the '(.+)' event using the '(.+)' API$/, asyn
             if (UTILS.getEnvVariable(CONSTANTS.CERTIFICATION) == true) {
               assert(false, `${CONSTANTS.PLATFORM_NOT_SUPPORT_LOG}: ${method}`);
             } else {
-              cy.log(`NotSupported: ${CONSTANTS.PLATFORM_NOT_SUPPORT_LOG}: ${method}`).then(() => {
-                throw new Error(CONSTANTS.STEP_IMPLEMENTATION_MISSING);
-              });
+              fireLog
```
parag-pv, line 334: Why is firelog and assert in new lines? If this was not done due to linting, a single line is more meaningful

cypress/support/e2e.js
```diff
@@ -51,9 +51,12 @@ Cypress.Commands.add('addContext', (context) => {
 Cypress.Commands.overwrite('log', (orig, string, options) => {
   orig(string, options);
   const dateString = new Date().toLocaleString().replace(',', '');
-  cy.addContext('[' + dateString + '][' + options + ']:[' + JSON.stringify(string) + ']');
+
+  typeof string == 'object' ? (string = JSON.stringify(string)) : string;
```
parag-pv, line 55: Minor semantic naming confusion here, `string == object`? That is 2 different datatypes. Could use a more meaningful name

cypress/support/cypress-support/src/utils.js
parag-pv, line 1: Can add an example for `firelog.info` usage at the top of the file

cypress/support/e2e.js
jmeier204: Yea "string" was the existing variable which we just removed the stringify operator off of and conditionally stringified only in case of it being an object type...i would say maybe we can name it something like "message" and just change all of the references inside the command overwrite to match.

cypress/support/step_definitions/fireboltCalls.js
Roebot926: This was from linting.

parag-pv (dismissed):

jmeier204 (approved):

PR #98, opened 2024-06-06
Systest 9779

kdivya153:


cypress/plugins/common.js
```diff
@@ -23,41 +23,40 @@ const CONSTANTS = require('../support/constants/constants');
```
kschrief, line 23: Are the changes to this file required?

cypress/support/cypress-support/src/FireboltInvoker.js
```diff
@@ -22,7 +22,7 @@ const logger = require('../../Logger')('FireboltInvoker.js');
 
 const WEBSOCKET = 'WebSocket';
 const DEFAULT_WS_URL_PROTOCOL = 'ws://';
-const DEFAULT_WS_PORT = 9998;
+const DEFAULT_WS_PORT = 8080;
```
kschrief, line 25: Has this been tested with Mock Firebolt?

cypress/plugins/common.js
kdivya153: if supportConfig.js doesn't exist it throwing an error and stop the execution.Particularly this change is required if there is no config module(default config module)

cypress/support/cypress-support/src/FireboltInvoker.js
kdivya153: Yes,I've tested in MFOS by executing suite runs.

PR #99, opened 2024-06-07
[FCS] Implement Setter/Getter runtime variables

Abhishk123:


cypress/support/cypress-support/src/utils.js
```diff
@@ -741,6 +741,72 @@ class FireLog {
...
+ * @description Resolve the device variable from the preprocessed data for the given key
+ * @example
+ * resolveAtRuntime(["result.{{attribute}}", "result.styles.{{attribute}}"])
+ * resolveAtRuntime("manage_closedcaptions.set{{attribute.uppercaseFirstChar}}")
+ * resolveAtRuntime("value")
+ *
+ * @returns
+ * ['result.fontSize', 'result.styles.fontSize']
+ * "manage_closedcaptions.setFontSize"
+ * "1.5"
```
jbigel, line 756: I think the code probably handles this just fine but if the value is set to `1.5`, it should return as a number not as a String

cypress/support/cypress-support/src/utils.js
```diff
@@ -741,6 +741,72 @@ class FireLog {
...
+    else if (!input.includes('{{')) {
+      return getEnvVariable(runtime).input || input;
+    }
+  } else if (Array.isArray(input) && input.length > 0) {
+    // input is an array; iterating through each element, it updates the actual value for that pattern if there is an occurrence of "{{".
+    return input.map((element) => {
+      if (element.includes('{{')) {
+        return replacingPatternOccurenceWithValue(element);
+      }
+    });
```
jbigel, line 804: Can we add some log if input is neither a String or an Array.

jbigel (changes requested):

cypress/support/cypress-support/src/utils.js
Abhishk123: it is returning the value present in rutime env as-is, this i missed to remove quote

cypress/support/cypress-support/src/utils.js
Abhishk123: addressed the comment

jbigel (approved):

neeradanelxsi (approved):

PR #102, opened 2024-06-07
RPPL-1875: Update fireboltCalls.js

Rajanika:
[RPPL-1875](https://ccp.sys.comcast.net/browse/RPPL-1875)

As part of the above ticket, we want to launch third party apps like Prime Video, Apple TV+, Hulu, etc. apart from FCA.
I observed that, by default, the discovery.launch() method launches the app with appId CONSTANTS.THIRD_PARTY_APP_ID.
Hence, with the proposed change, we are trying to update the CONSTANTS.THIRD_PARTY_APP_ID with any other appId, if present.
I have verified that it is working fine while launching FCA or any other app.


Rajanika:
Closing since these changes are not needed and taken care in 
https://github.com/comcast-firebolt/xumo-remote-config-module/pull/34
https://github.com/rdkcentral/firebolt-certification-suite/pull/109

PR #110, opened 2024-06-12
FIRECERT-2143: [FCS] Implement Setter/Getter 'set' Glue Code

Abhishk123:


cypress/support/cypress-commands/commands.js
```diff
@@ -35,15 +35,25 @@ Cypress.Commands.add(
...
     // Reading the data from combinedJson based on key.
     let fireboltData;
+    let logMessage = CONSTANTS.NO_DATA_FOR_THE_KEY + key;
     if (callType == CONSTANTS.SUPPORTED_CALLTYPES.FIREBOLTMOCKS) {
       fireboltData = UTILS.getEnvVariable(CONSTANTS.COMBINEDFIREBOLTMOCKS)[key];
     } else if (callType == CONSTANTS.SUPPORTED_CALLTYPES.SET_RESPONSE_JSON) {
       fireboltData = UTILS.getEnvVariable('setResponseJson')[key];
     } else {
-      fireboltData = UTILS.getEnvVariable(CONSTANTS.COMBINEDFIREBOLTCALLS)[key];
+      if (key) {
```
jbigel, line 44: I think this logic should be apart of the glue command. We don't want to set this runtime variable for non setter/getter calls.

Abhishk123: removed this changes from here and added in glue itself. will move it to some function in optimization ticket

cypress/support/step_definitions/fireboltCalls.js
```diff
@@ -422,3 +422,141 @@ Given(/User triggers event with value as '(.+)'/, (key) => {
...
+ * @function And '(.+)' invokes the '(.+)' API '(.+)' to set '(.+)' to '(.+)'
+ * @description send message to platform to make api call.
+ * @param {String} appId - app identifier, determines for which App sending the message to make a API call.
+ * @param {String} sdk - sdk name.
+ * @param {String} fireboltCallKey - key name passed to look for firebolt call object in fireboltCallData.
+ * @param {String} attribute - The attribute we are setting (ex. fontFamily).
+ * @param {String} value - The value we set the attribute to (ex. monospaced_sanserif)
+ * @example
+ * Given '1st party app' invokes the 'Firebolt' API 'CLOSEDCAPTION_SETTINGS' to set 'enable' to 'true'
+ * Given '3rd party app' invokes the 'Firebolt' API 'CLOSEDCAPTION_SETTINGS' to set 'enable' to 'true'
```
jbigel, line 437: 1st party update

cypress/support/step_definitions/fireboltCalls.js
```diff
@@ -422,3 +422,141 @@ Given(/User triggers event with value as '(.+)'/, (key) => {
...
+          UTILS.getEnvVariable('runtime').fireboltCall = fireboltData;
+        });
+      } else {
+        fireboltCallObject = UTILS.getEnvVariable('runtime').fireboltCall;
+        fireboltCallObjectErrorMessage =
+          'Unable to find the firebolt object in the runtime environment variable';
+      }
+
+      cy.then(() => {
+        if (!fireboltCallObject) {
```
jbigel, line 464: Can you add some descriptive comments

jbigel (changes requested):

cypress/support/step_definitions/fireboltCalls.js
```diff
@@ -422,3 +422,141 @@ Given(/User triggers event with value as '(.+)'/, (key) => {
     throw new Error(CONSTANTS.STEP_IMPLEMENTATION_MISSING);
   });
 });
+
+/**
+ * @module fireboltCalls
+ * @function And '(.+)' invokes the '(.+)' API '(.+)' to set '(.+)' to '(.+)'
```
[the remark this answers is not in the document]
Abhishk123: addressed the comment

cypress/support/step_definitions/fireboltCalls.js
Abhishk123: addressed the comment

cypress/support/step_definitions/fireboltCalls.js
Abhishk123: addressed the comment

cypress/support/cypress-support/src/utils.js
```diff
@@ -741,6 +741,38 @@ class FireLog {
...
+ *
+ * @returns
+ * 123
+ * true
+ */
+function parseValue(str) {
+  if (str === null || str === undefined) return str;
+
+  if (typeof str === 'string') {
+    if (str === 'true') return true;
```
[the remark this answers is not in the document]
Abhishk123: I don't think we use boolean value as string ex: "true", we will use boolean value always. In negative scenario we will use random string, something like "test"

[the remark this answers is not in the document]
neeradanelxsi: yes, we don't have a use case where we send **true** as a string.

jbigel (approved):

PR #114, opened 2024-06-14
Firelog does not work without child assertion function

bpvstaty366:


cypress/support/cypress-support/src/utils.js
```diff
@@ -658,27 +658,60 @@ global.resolveDeviceVariable = function (key) {
...
+        return this.log(...args);
+      }
+    `;
+    super('...args', functionBody);
+
+    const handler = {
+      apply: function (target, thisArg, argumentsList) {
+        let message;
+        const methodName = target.name;
+        if (methodName.toLowerCase().includes('equal'))
```
parag-pv, line 675: Having 'to be' just for `equal` might not be scalable. Since this condition is only for equal, what about assert.includes or assert.notIncludes or any new assertion? 

How about
if(more than 2 args) // for equal, include, deepEqual, strictEqual etc
message = argumentsList[argumentsList.length - 1] + 'Actual : ' + argumentsList[x] + ' Expected: ' + argumentsList[y]

else // for isNull, isDefined etc
message = argumentsList[argumentsList.length - 1] + 'Actual : ' + argumentsList[x]

bpvstaty366: updated, but I think for include and equal messages should be different 
I set for include - `["Error code -32400 expected to be present in the list of [-32602,-32400,-40300]"]`
for other methods with more than 2 args `["Method Content validation expected true to be true"]`

parag-pv: Can we make this into `Expected error code -32400 to be in [-32602,-32400,-40300]` for overall consistency with other logs?

cypress/support/cypress-support/src/utils.js
```diff
@@ -672,28 +672,72 @@ global.resolveDeviceVariable = function (key) {
...
-            return originalMethod.apply(this, args);
-          });
-        };
+      if (
+        method !== 'constructor' &&
+        method !== 'fireLog' &&
+        typeof instanceProxy[method] === 'function'
+      ) {
+        instanceProxy[method] = new Proxy(instanceProxy[method], handler);
+        if (['includes'].includes(method) || ['fail'].includes(method)) {
```
parag-pv, line 729: Do we need to specify each function that has it's own custom log in this if condition?
That will not be scalable. Can this be done dynamically?

cypress/support/cypress-support/src/utils.js
```diff
@@ -672,28 +672,72 @@ global.resolveDeviceVariable = function (key) {
...
+        let message;
+        const methodName = target.name;
+        if (target.hasOwnLog) {
+          // If the method has its own logging, just apply it
+          return Reflect.apply(target, thisArg, argumentsList);
+        } else {
+          if (argumentsList.length > 2)
+            message =
+              argumentsList[argumentsList.length - 1] +
+              ' expected ' +
```
parag-pv, line 696: Can the log be `Expected: <Expected> Actual: <Actual>`

There were concerns previously about users having difficulty differentiating between the actual response and source of truth.

cypress/support/constants/constants.js
```diff
@@ -91,6 +91,7 @@ module.exports = {
   EMAIL: 'email',
   ENV_PLATFORM_SDK_VERSION: 'platformSdkVersion',
   ERROR: 'error',
+  ERROR_CODE: 'Error code',
```
parag-pv, line 94: Is this being used? I don't see any reference to it in this PR.
If not used, please remove.

bpvstaty366: yes, it is used in 
`            fireLog.include(
              errorContentObject.errorCode,
              apiErrorResponse.code,
              CONSTANTS.ERROR_CODE
            );
`
that's why we had undefined error, ERROR_CODE wasn't added to constants

cypress/support/cypress-support/src/utils.js
bpvstaty366: added dynamic check

cypress/support/cypress-support/src/utils.js
bpvstaty366: fixed

parag-pv: I don't see this change in the PR, please confirm if the commit was pushed

bpvstaty366: yes, it wasn't commited, fixed it

parag-pv (approved):

ksentak (approved):

PR #117, opened 2024-06-20
FIRECERT 2155 - Add action from FCS to app/device for same state transitions

Nandana-NNR:
Ticket link - https://ccp.sys.comcast.net/browse/FIRECERT-2155

### DIFF
--- a/Scripts/lifeCycleAppObject.js
+++ b/Scripts/lifeCycleAppObject.js
@@ -97,8 +97,11 @@ class appConfig {
             this.history.push(this.state);
             logger.info('New appState pushed to history: ' + newState);
           }
-          if (!stateTransition.includes(newState)) {
+          if (!stateTransition.includes(newState) && currentState.state != newState) {
             cy.log('Requested state transition for application is not supported');
+            this.state = currentState;
+          }
+          if (currentState.state == newState) {
             this.state = currentState;
           }
         }
--- a/cypress/support/cypress-commands/lifecycle.js
+++ b/cypress/support/cypress-commands/lifecycle.js
@@ -362,12 +362,16 @@ Cypress.Commands.add('setAppState', (state, appId) => {
     // Set state to background
     case CONSTANTS.LIFECYCLE_STATES.BACKGROUND:
       // TODO: Checks for platform support
-      // Send message to platform to set app state to background
-      cy.setLifecycleState(state, appId).then(() => {
-        appObject.setAppObjectState(state);
-      });
+      if (
+        currentAppState.state != CONSTANTS.LIFECYCLE_STATES.BACKGROUND &&
+        currentAppState.state != CONSTANTS.LIFECYCLE_STATES.INITIALIZING
+      ) {
+        // If current app state is not background, send message to platform to set app state to background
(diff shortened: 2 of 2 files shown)


Scripts/lifeCycleAppObject.js
```diff
@@ -97,10 +97,13 @@ class appConfig {
...
           }
-          if (!stateTransition.includes(newState)) {
+          if (!stateTransition.includes(newState) && currentState.state != newState) {
             cy.log('Requested state transition for application is not supported');
             this.state = currentState;
           }
+          if (currentState.state == newState) {
+            this.state = currentState;
+          }
         }
```
neeradanelxsi, line 108: why do we need this change here ? expecting changes only inside **setAppState** function

cypress/support/cypress-commands/lifecycle.js
```diff
@@ -352,30 +352,20 @@ Cypress.Commands.add('setAppState', (state, appId) => {
...
-        // If current app state is not background, send message to platform to set app state to background
-        cy.setLifecycleState(state, appId).then(() => {
-          appObject.setAppObjectState(state);
-        });
-      }
+      // Send message to platform to set app state to background
+      cy.setLifecycleState(state, appId).then(() => {
+        appObject.setAppObjectState(state);
+      });
       break;
```
neeradanelxsi, line 370: No need to make any change to background as the api itself is not present in open rpc

cypress/support/cypress-commands/lifecycle.js
```diff
@@ -408,21 +398,19 @@ Cypress.Commands.add('setAppState', (state, appId) => {
...
       }
+      // Send lifecycle.suspend API call to 3rd party app
+      cy.invokeLifecycleApi(appId, CONSTANTS.LIFECYCLE_APIS.SUSPEND, {}).then((response) => {
+        if (response) {
+          cy.log(CONSTANTS.APP_RESPONSE + JSON.stringify(response));
+        }
+        appObject.setAppObjectState(state);
+        cy.lifecycleSchemaChecks(response, state);
+        // TODO: Checks for platform support
+      });
```
neeradanelxsi, line 413: No need to make any change to background as the api itself is not present in open rpc

Scripts/lifeCycleAppObject.js
Nandana-NNR: I think the above change is also required because the base logic here is if the newState is not present in 'allowedStateTransitions' config, then control comes to if condition in line 101, where it prints the log `cy.log('Requested state transition for application is not supported')` and assigns the state object to currentstate as there is no transition to newState. So since the same app transitions are also not part of 'allowedStateTransitions' config, say if u look at the relaunch scenario logs for example, you can see the same log. But ideally that's not correct right?
We can't say same state transition is not supported right as that could be misleading, so basically we are just  removing printing the log for same state transitions now that we have added action for it, rest is same. So its just a minor change to avoid printing the log and nothing major

neeradanelxsi (approved):

Nandana-NNR:
PR cancelled

PR #140, opened 2024-07-03
BREAKING CHANGE: FB interactions capturing

Nandana-NNR:


cypress/support/step_definitions/validations.md
```diff
@@ -28,18 +28,29 @@
 * `User validates lifecycle history for '1st party app' with 'background:foreground:background'`
 * `User validates lifecycle history for '3rd party app' with 'background:foreground'`
 
-## I '(start|stop)' performance metrics collection
+## Metrics collection process is '(initiated|stopped)'
 ### Purpose: To start or stop performance metrics service in device by passing appropriate intent to performance test handler
```
neeradanelxsi, line 33: query : why are we modifying performance glue code to handle metrics ?

cypress/support/step_definitions/validations.js
```diff
@@ -433,6 +433,41 @@ Given(
...
+            ? CONSTANTS.FAILED_TO_INITIATE_INTERACTIONS_SERVICE
+            : CONSTANTS.FAILED_TO_STOP_INTERACTIONS_SERVICE;
+        fireLog.assert(false, message);
+      }
+    });
+  } else {
+    cy.log(CONSTANTS.INTERACTIONS_SERVICE_ENABLED);
+  }
+});
+
```
neeradanelxsi, line 470: query : Is this glue code changes of FIRECERT-2190 ?

cypress/support/cypress-support/src/main.js
```diff
@@ -189,6 +196,13 @@ export default function (module) {
...
           });
         }
+        if (UTILS.getEnvVariable(CONSTANTS.IS_INTERACTIONS_SERVICE_ENABLED) == true) {
+          cy.startOrStopInteractionsService(CONSTANTS.STOPPED).then((response) => {
+            if (response) {
+              Cypress.env(CONSTANTS.IS_INTERACTIONS_SERVICE_ENABLED, false);
+            }
+          });
+        }
         // unsubscribing the list of topics
```
neeradanelxsi, line 206: is this befoerOperation  the equivalent of the glue code to start/stop fb interactions

cypress/support/step_definitions/validations.md
Eswar2103: It is md file, just updated the glue like in document for performance  metrics

cypress/support/step_definitions/validations.js
Eswar2103: not exactly, just added glue code in 2189, so can be used for testing bolt changes

cypress/support/cypress-support/src/main.js
Eswar2103: yes, you can use it in feature file or with before operation

cypress/support/cypress-commands/commands.js
```diff
@@ -539,6 +539,15 @@ Cypress.Commands.add('setResponse', (beforeOperation, scenarioName) => {
     cy.sendMessagetoPlatforms(requestMap).then((result) => {
       fireLog.isTrue(result.success, 'Response for marker creation: ' + JSON.stringify(result));
```
neeradanelxsi, line 540: can we add comment here

cypress/support/cypress-commands/commands.js
```diff
@@ -539,6 +539,15 @@ Cypress.Commands.add('setResponse', (beforeOperation, scenarioName) => {
...
     });
+  } else if (
+    beforeOperation.hasOwnProperty(CONSTANTS.INTERACTIONS_METRICS) &&
+    beforeOperation.interactionsMetrics === true
+  ) {
+    cy.startOrStopInteractionsService(CONSTANTS.INITIATED).then((response) => {
+      if (response) {
+        Cypress.env(CONSTANTS.IS_INTERACTIONS_SERVICE_ENABLED, true);
+      }
+    });
```
neeradanelxsi, line 550: lets make sure that any error which occurs here is handled properly and not cause the rest of the tests to fail/skipped

cypress/support/cypress-support/src/main.js
```diff
@@ -55,6 +55,22 @@ export default function (module) {
       } else {
         cy.log('Unable to establish a pub/sub connection.');
       }
```
neeradanelxsi, line 57: comment to be added here

cypress/support/cypress-support/src/main.js
```diff
@@ -55,6 +55,22 @@ export default function (module) {
...
+      const topic = UTILS.getTopic(
+        UTILS.getEnvVariable(CONSTANTS.FIRST_PARTY_APPID),
+        CONSTANTS.SUBSCRIBE,
+        null,
+        CONSTANTS.TOPIC_FBINTERACTIONS
```
neeradanelxsi, line 62: I am not sure I understood line 65 . what is that trying to do >?

cypress/support/cypress-support/src/main.js
```diff
@@ -55,6 +55,22 @@ export default function (module) {
...
+      appTransport.subscribe(topic, UTILS.interactionResults);
+      if (UTILS.getEnvVariable(CONSTANTS.INTERACTIONS_METRICS, false) == true) {
+        cy.startOrStopInteractionsService(CONSTANTS.INITIATED).then((response) => {
+          if (response) {
+            Cypress.env(CONSTANTS.IS_INTERACTIONS_SERVICE_ENABLED, true);
+          }
+        });
+      } else {
+        cy.log(CONSTANTS.INTERACTIONS_SERVICE_NOT_ACTIVE);
+      }
```
neeradanelxsi, line 73: what is the difference between the interaction service being active and enabled ? 
what if there is no response in line 67

cypress/support/cypress-support/src/main.js
```diff
@@ -189,6 +204,13 @@ export default function (module) {
             }
           });
         }
+        if (UTILS.getEnvVariable(CONSTANTS.IS_INTERACTIONS_SERVICE_ENABLED, false) == true) {
+          cy.startOrStopInteractionsService(CONSTANTS.STOPPED).then((response) => {
+            if (response) {
+              Cypress.env(CONSTANTS.IS_INTERACTIONS_SERVICE_ENABLED, false);
+            }
+          });
+        }
```
neeradanelxsi, line 213: I am not sure I understand what line 207 means . and what is 210 trying to do

cypress/support/cypress-commands/commands.js
Abhishk123: added the comment

cypress/support/cypress-commands/commands.js
Abhishk123: Sure

cypress/support/cypress-support/src/main.js
Abhishk123: Addressed the comment

cypress/support/cypress-support/src/main.js
Abhishk123: initiating the Interaction service to get the logs

cypress/support/cypress-support/src/main.js
Abhishk123: here we are stopping the service if this enabled in before block

cypress/support/cypress-commands/commands.js
kevinshahfws: We should do PoC on printing the before operation logs. Silently running the rest of execution might cause complex debug problems.

cypress/support/step_definitions/validations.js
```diff
@@ -277,3 +277,54 @@ Given(
...
+/**
+ * @module validations
+ * @function Interactions metrics collection process is (initiated|stopped)
+ * @description To start or stop listening to firebolt interactions in device by passing appropriate intent to designated handler
+ * @param {String} action - initiated or stopped
+ * @example
+ * Given Interactions metrics collection process is initiated
+ * Given Interactions metrics collection process is stopped
+ */
+Given(/Interactions metrics collection process is (initiated|stopped)/, (action) => {
```
kevinshahfws, line 290: I would remove metrics from this line.

neeradanelxsi (dismissed):

README.md
```diff
@@ -398,6 +398,48 @@ HTTP call to the platform:<br/>
   }
   ```
 
+## Interaction Log service
```
jbigel, line 401: Can this go in a separate markdown file somewhere or at the very least create a table of contents for the main doc? Our main readme is becoming quite large and without any quick links it just looks daunting rather than helpful

cypress/support/step_definitions/validations.js
```diff
@@ -279,3 +279,54 @@ Given(
...
+/**
+ * @module validations
+ * @function Given Validate Firebolt Interactions logs
+ * @description Validating the firebolt interaction logs in configModule
+
+ * @example
+ * Given Validate Firebolt Interactions logs
+ */
+Given(/Validate Firebolt Interactions logs/, () => {
+  cy.then(() => cy.validateFireboltInteractionLogs());
```
jbigel, line 331: Is there a reason this test step just has one line pointing to a command that only has 5 lines itself? Not sure if there some cypress/JS caveat forcing this but reading it seems those 5 lines could just be put here instead of unnecessarily pointing to some command in another file?

jbigel (changes requested):

cypress/support/step_definitions/validations.js
Eswar2103: This interaction logs validation is up to the config module owners, we just gave them an ability to add their own validation for logs, They can add their own custom validation function in config module and this glue will do the remaining work.

README.md
Eswar2103: Moved interaction logs readme to a new md file, added that new file inside DOCS

kevinshahfws (dismissed):

Docs/Interaction_Logs.md
```diff
@@ -0,0 +1,42 @@
+## Interaction Log service
+### Background -
+The interaction log service will notify all the Firebolt calls getting invoked in the platform whether the Firebolt call is made manually by any action or internally made by any application or service. Whenever the platform identifies a firebolt call, the interaction log service will notify that.
+### Uses -
+The Interaction log service is essential for tracking method calls. When a method is called, it internally triggers/invokes other Firebolt methods. This service can be used to obtain the list of internal method calls.
```
neeradanelxsi, line 5: Interaction log service is used to capture the various firebolt calls invoked during a test execution. This service will notify all the Firebolt calls getting invoked in the platform if the Firebolt call is called manually by any action or internally called by any application or service.

neeradanelxsi: Uses 
The Interaction log service is essential for tracking all the firebolt calls made by the platform. When a method is called, it may internally trigger/invoke other Firebolt methods. This service can be used to obtain the list of internal method calls and validate them if needed.

Docs/Interaction_Logs.md
```diff
@@ -0,0 +1,42 @@
...
+### Configurations -
+#### Framework side - fcs/config module side
+Below are the different ways to start the interaction service
+* Using `interactionsMetrics` flag. This can be enabled from the cli command to start the service
+* Using glue codes we can able to start and stop the interaction service and then using validation glue able to validate the obtained logs.
+* Using Before operation, able to start the service by adding `interactionsMetrics` to true in [beforeOperation](#before-operation). 
+
```
neeradanelxsi, line 12: We should mention about Pubsub here . FCS would be listening to fb_interactions topic right ?

neeradanelxsi: * Using `interactionsMetrics` flag. This can be enabled from the cli command to start the service
* Using glue codes we can able to start and stop the interaction service and then using validation glue able to validate the obtained logs.
* Using Before operation, able to start the service by adding `interactionsMetrics` to true in [beforeOperation](#before-operation). 

to be changed to below
* Using `interactionsMetrics` flag. By default, this is false. This can be enabled from the cli command to start the service
* Using glue codes, we can start and stop the interaction service for capturing the logs and then using validation glue, validate these logs. **_List down the glue codes we have for this. And provide link to the readme_**
* Using Before operation, interaction service can be started by adding `interactionsMetrics` to true in [beforeOperation](#before-operation). **Provide a sample beforeOperation object as reference**

Docs/Interaction_Logs.md
```diff
@@ -0,0 +1,42 @@
...
+#### Device side -
+Platform should have the support of interactonaction logging service. The way interactionlogs will be fetched from device differs on platform and that can be handled through config module.
+## Implementations -
```
neeradanelxsi, line 15: interactonaction  - this is a typo

Device side - I believe we should also mention about fb interaction here

Docs/Interaction_Logs.md
neeradanelxsi: I think you have added all that below. about point 2 and 3. You can keep it there and provide reference here . or you can move all those up also

Docs/Interaction_Logs.md
Eswar2103: done

Docs/Interaction_Logs.md
Eswar2103: done

Docs/Interaction_Logs.md
Eswar2103: done

jbigel (dismissed):

neeradanelxsi (dismissed):

Abhishk123 (approved):

neeradanelxsi (approved):
Re-approving after pulling latest dev into the branch

PR #134, opened 2024-06-28
FIRECERT:2171 - Ported testcase for device, display and MediaInfo.

SnehalP251:

PR #145, opened 2024-07-10
User triggers event implementation

Nandana-NNR:


cypress/support/constants/constants.js
```diff
@@ -325,6 +325,7 @@ module.exports = {
     RECORD_LIFECYCLE_HISTORY: 'fcs.recordLifecycleHistory',
     SETPERFORMANCETESTHANDLER: 'performance.setPerformanceTestHandler',
     SETLIFECYCLESTATE: 'fcs.setLifecycleState',
+    TRIGGEREVENT: 'fcs.triggerEvent',
```
Abhishk123, line 328: add this override function name to REQUEST_OVERRIDE_CALLS in constants

cypress/support/step_definitions/fireboltCalls.js
```diff
@@ -527,15 +527,23 @@ Given(
...
+ * @description sending message to platform to make post call to set event values.
+ * @param {String} key - key name of the event data
  * @example
- * And User triggers event with value as ' DEVICE_ONHDCPCHANGED_EVENTS'
+ * And User triggers event with value as 'onColorimetryChanged event with colorimetry as BT2020cYCC'
  */
 Given(/User triggers event with value as '(.+)'/, (key) => {
-  fireLog.info(CONSTANTS.STEP_DEFINITION_NEEDS_TO_IMPLEMENT).then(() => {
-    throw new Error(CONSTANTS.STEP_IMPLEMENTATION_MISSING);
+  const requestMap = {
```
Abhishk123, line 537: any reason why we not added the logic to fetching the key value here?

Nandana-NNR: Addressed

cypress/support/constants/constants.js
Nandana-NNR: This is added in REQUEST_OVERRIDE_CALLS only

cypress/support/constants/constants.js
```diff
@@ -457,6 +458,10 @@ module.exports = {
   SECONDARY_THIRD_PARTY_APP_ID: 'secondary3rdPartyAppId',
   SECONDARY_APPID_MISSING_ERROR:
     '`Unable to find the ${envAppIdKey} value in the env, please add the value in configModule/constants/config.json`',
+  TRIGGER_EVENT_MISSING:
+    'triggerEvent request override function to set event values not implemented by underlying platform',
+  TRIGGER_EVENT_REQUEST: 'Request sent to platform to trigger event: ',
+  TRIGGER_EVENT_SUCCESS: 'Event triggered successfully',
 };
 function getSanityReportPath() {
```
neeradanelxsi, line 466: Log line for 461 does not make sense 
Line 463 - Log doesnt make sense . In this case who is triggering the event ? Has there been a case where events are sent to the platform  ?

cypress/support/step_definitions/fireboltCalls.js
```diff
@@ -527,15 +527,26 @@ Given(
...
- * @function User triggers event with value '{}}'
- * @description sending message to platform to make post call to set values.
- * @param {String} key - Name of event to be called.
+ * @function User triggers event with value as '(.+)'
+ * @description sending message to platform to make post call to set event values.
+ * @param {String} key - key name of the event data
  * @example
- * And User triggers event with value as ' DEVICE_ONHDCPCHANGED_EVENTS'
+ * And User triggers event with value as 'onColorimetryChanged event with colorimetry as BT2020cYCC'
  */
```
neeradanelxsi, line 535: can you give an example of actual FB event ?

cypress/TestCases/FireboltCertification/Device.feature
```diff
@@ -51,7 +51,7 @@ Feature: Device
         Then 'Firebolt' platform responds to '1st party app' for 'set device name to kitchen'
         And 'Firebolt' platform does not trigger event for 'onDeviceNameChanged'
 
-    @Device @coreSDK @sdk @transport @notSupported
+    @Device @coreSDK @sdk @transport @requiresPlatformImplementation
```
neeradanelxsi, line 54: don't remove @notSupported tag for TP test cases. those are still required along with requiresPlatformImplementation tag

neeradanelxsi (approved):

cypress/support/step_definitions/fireboltCalls.js
```diff
@@ -527,15 +527,26 @@ Given(
...
+ * @param {String} key - key name of the event data
  * @example
- * And User triggers event with value as ' DEVICE_ONHDCPCHANGED_EVENTS'
+ * And User triggers event with value as 'onNetworkChanged events with wifi connected'
  */
 Given(/User triggers event with value as '(.+)'/, (key) => {
-  fireLog.info(CONSTANTS.STEP_DEFINITION_NEEDS_TO_IMPLEMENT).then(() => {
-    throw new Error(CONSTANTS.STEP_IMPLEMENTATION_MISSING);
+  cy.fireboltDataParser(key).then((parsedData) => {
+    const value = parsedData[0].params;
```
jbigel, line 538: Curious about the data being an array and statically taking the first value. Can you explain this?

Nandana-NNR: This is because current fireboltDataParser implementation returns data as an array and not object. Also in our case since we are setting event value for a given event passed in key, the array will only have 1 data object and hence the code

The basic intention of the ticket was to unblock Snehal's ticket FIRECERT-2171 which needs this glue support. All the json objects are added in fireboltCalls as part of this ticket (https://github.com/rdkcentral/firebolt-certification-suite/pull/134/files), and here I am trying to parse through that data and set values. I remember giving a try to move a value to fireboltMocks and use getFireboltData() instead, but I was facing errors in that case. So the decision was to go ahead with fireboltDataParser as the latter is outside the scope of this ticket and might need additional work if to be taken as a seperate ticket

jbigel: Understood, thank you for elaborating.

jbigel (dismissed):

Abhishk123 (approved):

PR #147, opened 2024-07-11
[FCS] Setter/Getter Code Refactoring

Abhishk123:


cypress/support/cypress-commands/commands.js
```diff
@@ -833,3 +834,294 @@ Cypress.Commands.add('clearCache', () => {
...
+    } else {
+      fireLog.info(
+        `${target} returned response in invalid format, which could lead to failures in validations. Response must be in JSON RPC format - ${response}`
+      );
+    }
+  });
+});
+
+Cypress.Commands.add('methodorEventResponseValidation', (validationType, additionalParams) => {
+  const { method, context, contentObject, expectingError, appId, eventExpected } = additionalParams;
```
neeradanelxsi, line 967: the new commands look good. 
methodorEventResponseValidation and sendMessageToPlatformOrApp

cypress/support/step_definitions/getterSetterCalls.js
```diff
@@ -16,25 +16,512 @@
...
+        } else {
+          cy.log(
+            `${CONSTANTS.SKIPCONTENTVALIDATION} flag is enabled, Skipping the Content validation`
+          );
+        }
+      });
+    } else {
+      assert(false, `${sdk} SDK not Supported`);
+    }
+  }
```
neeradanelxsi, line 526: regarding the other code inside each glue code (for eg, data parsing, request map creation etc, ) , is it possible to combine that also to functions and reuse in multiple glue codes ?
May not be possible, but if it is a possibility, it would be really good

Abhishk123: In existing glue code optimised and for new glue code having different logic there so kept as it is

cypress/TestCases/FireboltCertification/Accessibility.feature
```diff
@@ -1,157 +1,155 @@
...
+    Scenario Outline: Accessibility.closedCaptions - Positive Scenario: <Scenario>
+        Given we test the 'CLOSEDCAPTIONS' getters and setters '<Method>' to '<Value>'
+        When '3rd party app' registers for the 'Firebolt' event
+        Given 1st party app invokes the 'Firebolt' API to set '<Method>' to '<Value>'
+        And 'Firebolt' platform responds to '1st party app' set API
+        When '3rd party app' invokes the 'Firebolt' get API
+        And 'Firebolt' platform responds to '3rd party app' get API
+        And 'Firebolt' platform triggers '3rd party app' event
 
-   @Accessibility @coreSDK @sdk @transport @requiresPlatformImplementation
```
jbigel, line 111: Do we not need a clear listeners scenario anymore?

cypress/TestCases/FireboltCertification/Accessibility.feature
```diff
@@ -1,157 +1,155 @@
...
-      When '3rd party app' registers for the 'accessibility onClosedCaptionsSettingsChanged' event using the 'Firebolt' API
-      And '3rd party app' invokes the 'Firebolt' API to 'get closedCaptions settings'
-      And 1st party app invokes the 'Firebolt' API to '<Key>'
-      Then 'Firebolt' platform responds to '1st party app' for '<Key>'
-      When '3rd party app' invokes the 'Firebolt' API to 'get closedCaptions settings'
-      Then 'Firebolt' platform responds with '<Method_Content>'
-      And 'Firebolt' platform triggers event '<Event_Content>'
+    @Accessibility @coreSDK @sdk @transport
+    Scenario Outline: Accessibility.closedCaptionsSettings - Positive Scenario: <Scenario> with 'null' params
+        Given we test the 'CLOSEDCAPTIONS_SETTINGS' getters and setters '<Method>' to '<Value>'
```
jbigel, line 55: We reference `Value` but that doesn't exist in the examples?

cypress/TestCases/FireboltCertification/Manage/ClosedCaptionsManage.feature
```diff
@@ -2,113 +2,114 @@ Feature: ClosedCaptions_Manage
...
     @ClosedCaptions @manageSDK
-    Scenario Outline: Closedcaptions.<Method> - Positive Scenario: <Scenario> with 'null' params
-        When 1st party app registers for the '<Event>' event using the 'Firebolt' API
-        And 1st party app invokes the 'Firebolt' API to '<Set_API_Key>'
-        Then 'Firebolt' platform responds to '1st party app' for '<Set_API_Key>'
-        When 1st party app invokes the 'Firebolt' API to '<API_Key>'
-        Then 'Firebolt' platform responds to '1st party app' with '<Method_Validation_Key>'
-        And 'Firebolt' platform triggers to '1st party app' with '<Event_Validation_Key>'
+    Scenario Outline: ClosedCaptions.<Method> - Positive Scenario: <Scenario>
+        Given we test the 'CLOSED_CAPTIONS_SETTINGS' getters and setters '<Method>' to '<Value>'
```
jbigel, line 52: We reference Value but that doesn't exist in the examples?

cypress/support/cypress-commands/commands.js
jbigel, line 1: I don't see any issues with the code but just wondering if these commands could go in a file that is more indicative of their purpose? I know we have a validations folder, could the validation command go somewhere under there? 
 This may be something we don't want to address as a part of this PR but in a ticket for project wide re-structuring of the code organization.

cypress/support/step_definitions/getterSetterCalls.js
```diff
@@ -16,25 +16,509 @@
...
+  async (key, attribute, value) => {
+    // Clear any current env.runtime variables
+    Cypress.env(CONSTANTS.RUNTIME, {});
+
+    // Look for the firebolt call key in firebolt data
+    cy.getFireboltData(key).then((parsedData) => {
+      // getFireboltData handles the case where the key is not found
+      // Save the object as env.runtime.fireboltCall
+      const runtime = { fireboltCall: parsedData };
+      if (attribute && value) {
```
jbigel, line 42: So in conjunction with the comment on the null scenarios from the feature file, do we want the ability to just provide `attribute`?

cypress/support/step_definitions/getterSetterCalls.js
```diff
@@ -16,25 +16,509 @@
...
+ * @param {String} appId - app identifier.
+ * @param {String} sdk - sdk name.
+ * @param {String} fireboltCallKey - key name passed to look for firebolt call object in fireboltCallData.
+ * @example
+ * Given '1st party app' invokes the 'Firebolt' get API 'CLOSEDCAPTION_SETTINGS'
+ * Given '3rd party app' invokes the 'Firebolt' get API 'CLOSEDCAPTION_SETTINGS'
+ * Given '3rd party app' invokes the 'Firebolt' get API
+ * Given 'test_app' invokes the 'Firebolt' get API 'CLOSEDCAPTION_SETTINGS'
+ */
+Given(/'(.+)' invokes the '(.+)' get API(?: '(.+)')?$/, async (appId, sdk, fireboltCallKey) => {
```
jbigel, line 173: With the exception of the `appId` logic this code seems very similar to the previous one. Can we extract common functionality into a helper method to be used by both?

cypress/support/step_definitions/getterSetterCalls.js
```diff
@@ -16,25 +16,509 @@
...
+ * @description Sending a message to platform or app to register an event
+ * @param {String} appId - app identifier.
+ * @param {String} sdk - sdk name.
+ * @param {String} fireboltCallKey - key name passed to look for firebolt call object in fireboltCallData.
+ * @example
+ * And '1st party app' registers for the 'Firebolt' 'CLOSEDCAPTION_SETTINGS' event
+ * And '3rd party app' registers for the 'Firebolt' 'CLOSEDCAPTION_SETTINGS' event
+ * And '1st party app' registers for the 'Firebolt' event
+ */
+Given(
```
jbigel, line 266: I think this one shares a good amount of logic as well, ideally we could create a helper function or two if needed to so that they all use the same logic?

cypress/support/step_definitions/getterSetterCalls.js
```diff
@@ -16,25 +16,509 @@
...
+ * @param {String} errorContent - Doing error content validation when error content object key passed. Ex: 'INVALID_TYPE_PARAMS'
+ * @example
+ * And 'Firebolt' platform triggers '1st party app' 'CLOSEDCAPTION_SETTINGS' event
+ * And 'Firebolt' platform triggers '1st party app' event
+ * And 'Firebolt' platform triggers '3rd party app' 'CLOSEDCAPTION_SETTINGS' event
+ * And 'Firebolt' platform does not trigger '3rd party app' 'CLOSEDCAPTION_SETTINGS' event
+ * And 'Firebolt' platform triggers '1st party app' event
+ * And 'Firebolt' platform triggers '1st party app' 'CLOSEDCAPTION_SETTINGS' event with 'INVALID_TYPE_PARAMS'
+ */
+Given(
```
jbigel, line 459: Wondering if this one and the one above could also use the a single helper function?

jbigel (changes requested):

cypress/TestCases/FireboltCertification/Manage/ClosedCaptionsManage.feature
Abhishk123: update the step adding the value as null

cypress/support/step_definitions/getterSetterCalls.js
Abhishk123: Yes, we have some scenarios where we passing null value for set calls. I think it should be better if we have that support.
@neeradanelxsi

cypress/support/cypress-commands/commands.js
Abhishk123: @neeradanelxsi

cypress/support/step_definitions/getterSetterCalls.js
Abhishk123: In upcoming enhancement ticket observed we are going to modify the glue and removing some fields, After removing some of the lines we need to remove. Is it to handle thins in new ticket?

cypress/support/step_definitions/getterSetterCalls.js
Abhishk123: In upcoming enhancement ticket observed we are going to modify the glue and removing some fields, After removing some of the lines we need to remove. Is it to handle thins in new ticket?

cypress/support/step_definitions/getterSetterCalls.js
Abhishk123: In upcoming enhancement ticket observed we are going to modify the glue and removing some fields, After removing some of the lines we need to remove. Is it to handle thins in new ticket?

jbigel (dismissed):

cypress/TestCases/FireboltCertification/Accessibility.feature
Abhishk123: Currently clear listeners glue for v2 not there, will take care in new ticket

anjalimukundan (approved):

PR #150, opened 2024-07-17
Integrated player testcase

Nandana-NNR:


cypress/TestCases/FireboltCertification/AdditionalTestcases/IntegratedPlayer.feature
```diff
@@ -0,0 +1,36 @@
...
+
+        Examples:
+            | Scenario                    | API_Key                              | Validation_key                   | method         |
+            | Validate create api         | create broadcastPlayer instance      | broadcastPlayer instance         | create         |
+            | Validate status api         | get status of broadcastPlayer        | status of broadcastPlayer        | status         |
+            | Validate provideStatus api  | get broadcast status of player       | broadcast status of player       | provideStatus  |
+            | Validate createResponse api | get response from createprovider     | response of createprovider       | createResponse |
+            | Validate createError api    | get error from createprovider        | error of createprovider          | createError    |
+            | Validate statusResponse api | get response from statusprovider     | response of statusprovider       | statusResponse |
+            | Validate statusError api    | get error from statusprovider        | error of statusprovider          | statusError    |
```
neeradanelxsi, line 36: not sure I understand these test cases . Are you trying to call rpc-only methods here ? 


PlayerApp creates the Player based on the first IUI request ... Based on the type either Broadcast or Streaming Player gets created...
Then next call from IUI would be load/play content which will be performed in the Player created by above step

cypress/TestCases/FireboltCertification/AdditionalTestcases/IntegratedPlayer.feature
```diff
@@ -0,0 +1,64 @@
+Feature: IntegratedPlayer
+
+    @initialization
+    Scenario: Launch FCA for 'IntegratedPlayer'
+        Given the environment has been set up for 'IntegratedPlayer' tests
+        And 3rd party 'certification' app is launched with 'inactive' state
```
neeradanelxsi, line 6: we need the test case to be separate here. 
Creating player can be a single test case
Then all the player method calls like load, play etc should be separate. If we try to keep all of them together in one single test case, if one validation fails, none of the other methods will be validated.

cypress/TestCases/FireboltCertification/AdditionalTestcases/IntegratedPlayer.feature
```diff
@@ -0,0 +1,64 @@
...
+        Then 'Firebolt' platform responds to '1st party app' with 'value for provide status'
+        When 1st party app invokes the 'Firebolt' API to 'get player status'
+        Then 'Firebolt' platform responds to '1st party app' with 'value for player status'
+        When 1st party app invokes the 'Firebolt' API to 'get broadcast status of player'
+        Then 'Firebolt' platform responds to '1st party app' with 'broadcast status of player'
+        When 1st party app invokes the 'Firebolt' API to 'stop player'
+        Then 'Firebolt' platform responds to '1st party app' with 'value for player stop'
+        When 1st party app invokes the 'Firebolt' API to 'destroy player'
+        Then 'Firebolt' platform responds to '1st party app' with 'value for player destroy'
+
```
neeradanelxsi, line 38: also, regarding the 3rd flow , check with kevin on the event name
also the difference in methods for streaming player and broadcast player,  mention that to kevin. Maybe it was a miss or it was intentional

cypress/TestCases/FireboltCertification/AdditionalTestcases/IntegratedPlayer.feature
```diff
@@ -0,0 +1,64 @@
...
+    @initialization
+    Scenario: Launch FCA for 'IntegratedPlayer'
+        Given the environment has been set up for 'IntegratedPlayer' tests
+        And 3rd party 'certification' app is launched with 'inactive' state
+
+    @sdk @IntegratedPlayer
+    Scenario: Positive Scenario : Create and validate broadcast player instance
+        When 1st party app invokes the 'Firebolt' API to 'create broadcastPlayer instance'
+        Then 'Firebolt' platform responds to '1st party app' with 'broadcastPlayer instance'
+        When 1st party app invokes the 'Firebolt' API to 'get status of broadcastPlayer'
```
neeradanelxsi, line 12: also we can place TC first before creating a player . 

- So suppose if we call player.load, player.play without creating a broadcast/streaming player, will the platform throw error ?

or what does player.destroy do ? 
will it destroy the player ?  can you confirm that?

cypress/TestCases/FireboltCertification/AdditionalTestcases/IntegratedPlayer.feature
```diff
@@ -0,0 +1,64 @@
+Feature: IntegratedPlayer
+
+    @initialization
+    Scenario: Launch FCA for 'IntegratedPlayer'
+        Given the environment has been set up for 'IntegratedPlayer' tests
+        And 3rd party 'certification' app is launched with 'inactive' state
+
+    @sdk @IntegratedPlayer
+    Scenario: Positive Scenario : Create and validate broadcast player instance
```
neeradanelxsi, line 9: we would need to write the tc in xumo and mf config module.

cypress/TestCases/FireboltCertification/AdditionalTestcases/IntegratedPlayer.feature
Nandana-NNR: api and behavorial testcases seperated with player creation as a single test

cypress/TestCases/FireboltCertification/AdditionalTestcases/IntegratedPlayer.feature
Nandana-NNR: ported to config module

cypress/support/constants/constants.js
```diff
@@ -303,6 +303,7 @@ module.exports = {
   PLATFORM_NOT_SUPPORT_LOG: 'Platform does not support method',
   PLATFORM_NOT_TRIGGER_EVENT: 'Platform MUST not trigger event ',
   PLATFORM_TRIGGER_EVENT: 'Platform MUST trigger event ',
+  PLAYERID: 'playerId',
   PREREQUISITE_DATA: 'PreRequisiteData.json',
```
neeradanelxsi, line 307: not sure why we keep adding player in FCS. This has to be removed

Nandana-NNR: removed

PR #161, opened 2024-07-24
fixErrorMessages : Fixing error messages

neeradanelxsi:


cypress/plugins/testDataProcessor.js
```diff
@@ -102,45 +102,22 @@ function processSetResponseJson(setResponseJsonData) {
 }
 
 /**
- *  @function processErrorContentJson
- *  processErrorContentJson function will perform following operations
+ *  @function getErrorContentObjectJson
+ *  getErrorContentObjectJson function will perform following operations
  *  - Iterate over each key in the provided JSON
```
jbigel, line 107: This description no longer reflects what is happening in the function.

cypress/plugins/testDataProcessor.md
```diff
@@ -44,9 +44,9 @@ Response
 
 - {'abc': {'method': 'method_name', 'validationJsonPath': 'result', 'content': true}}
 ```
-
-### > processErrorContentJson
-processErrorContentJson function will perform following operations
+getErrorContentObjectJson
+### > 
+getErrorContentObjectJson function will perform following operations
```
jbigel, line 49: same as above.

jbigel (approved):
Approving to unblock reports but would like to see changes.

Ideally I'd like to see FCS define the different error types and config modules can set what error codes correspond to the various types. Then in the testDataProcessor we can fetch the error codes from the config mod and plug it into our expected error types.

cypress/plugins/testDataProcessor.js
Abhishk123: addressed the comment

cypress/plugins/testDataProcessor.md
Abhishk123: addressed the comment

jbigel (approved):

PR #182, opened 2024-08-02
Added document for error validation

Abhishk123:


cypress/support/constants/constants.js
```diff
@@ -321,7 +321,6 @@ module.exports = {
   PLATFORM_NOT_TRIGGER_EVENT: 'Platform MUST not trigger event ',
   PLATFORM_TRIGGER_EVENT: 'Platform MUST trigger event ',
   PLATFORM_RELEASE: 'Release',
-  PLATFORM: 'Platform',
   PARTNER: 'Partner',
```
neeradanelxsi, line 324: I think this got fixed in https://github.com/rdkcentral/firebolt-certification-suite/pull/179 pr

cypress/support/step_definitions/validations.md
```diff
@@ -412,3 +445,156 @@ While validating, if a key is present in both fcs-validation jsons (eg: cypress/
             }
         ]
 }
+
+
```
neeradanelxsi, line 449: Change Background into 

There are some test cases which involves calling firebolt methods with invalid parameters, missing parameters. Here, we are expecting error to be returned from the device. FCS has some default error types defined which can be used in various test cases. 

Default Validation Types in FCS
Below is the default error objects supported in FCS [errorContentObjects.json](https://github.com/rdkcentral/firebolt-certification-suite/blob/e97116c88ba79a2ad7f2fe23d55dc6101c3dd484/cypress/fixtures/objects/errorContentObjects.json)

NOT_SUPPORTED - This can be used for api's which are not supported
NOT_PERMITTED - This can be used for api's which are not supported
NOT_AVAILABLE - This can be used for api's which are not supported
INVALID_TYPE_PARAMS - This can be used for error validation which involves calling firebolt methods with invalid parameters, missing parameters etc.

neeradanelxsi: Remove the note

cypress/support/step_definitions/validations.md
```diff
@@ -412,3 +445,156 @@ While validating, if a key is present in both fcs-validation jsons (eg: cypress/
...
+```
+    "NOT_SUPPORTED": {
+        "type": "schemaOnly"
+    }
+```
+
+Note:
+
+- Error content validation will be happened only when override objects are added from the config module.
+- If error type having `schemaOnly` as shown in example, exection will stop after schema validation and testcase will PASS.
```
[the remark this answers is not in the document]
neeradanelxsi: Replace below 
Validation objects can be added in config module: cypress/fixtures/objects/errorContentObjects.json file.
If an object in the config module shares the same key name as an error object defined in FCS, FCS error object will be overriden.

with 

-  User can add error code validations to error validations by overriding in config module in the cypress/fixtures/objects/errorContentObjects.json file.

- For that, they would need to use the same key name as any of the default error objects defined in FCS.

Format 
Below is the format need to be followed while adding override in config module

{
    "INVALID_TYPE_PARAMS": {
        "type": "errorValidationFunction",
        "validations": [
            {
                "type": {
                    "errorCode": [ERROR_CODE1, ERROR_CODE2]
                }
            }
        ]
    }
}

Note: FCS expects the error object to be defined in the above format. Any deviation from the above format would cause failures in error content validation.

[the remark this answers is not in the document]
neeradanelxsi: Replace **Any additional validations needed, that can be done using custom validation** with **Custom validation support**

Then add details as shown below
Users can further add their own error validations using custom validation objects. 

- To do custom validation, error object must have the type as custom shown in below format.
- The value given for assertionDef represents the function name and this function should be defined in config module: cypress/fixtures/customValidations/

cypress/support/step_definitions/validations.md
Abhishk123: addressed the comment

cypress/support/step_definitions/validations.md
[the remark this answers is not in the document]
Abhishk123: addressed the comment

cypress/support/constants/constants.js
Abhishk123: yes, removed this change

neeradanelxsi (approved):

neeradanelxsi (approved):

PR #188, opened 2024-08-06
feat: added support in env setup glue for fetching device details dynamically

anjalimukundan:


cypress/support/step_definitions/testSetup.js
```diff
@@ -55,6 +55,31 @@ Given('the environment has been set up for {string} tests', (test) => {
...
+            }
+            Cypress.env(CONSTANTS.DEVICE_DATA, extractedData);
+          } else {
+            throw new Error('Error getting deviceData from CodeBig ');
+          }
+        } else {
+          throw new Error('Error getting device id ');
+        }
+      }
+    );
```
neeradanelxsi, line 82: codebig is something proprietary. 
Also, this would fail for other config modules, right? Shouldn't this stay in config module before operation?

anjalimukundan: moved the code to configModule and configHelper

cypress/support/step_definitions/testSetup.js
```diff
@@ -55,6 +55,21 @@ Given('the environment has been set up for {string} tests', (test) => {
     if (Cypress.env(CONSTANTS.TEST_TYPE).includes('rpc-Only')) {
       Cypress.env(CONSTANTS.IS_RPC_ONLY, true);
     }
+    // fetch device details dynamically
```
neeradanelxsi, line 58: this should be a list as mentioned yesterday

cypress/support/step_definitions/testSetup.js
```diff
@@ -55,6 +55,21 @@ Given('the environment has been set up for {string} tests', (test) => {
...
+            const method = CONSTANTS.REQUEST_OVERRIDE_CALLS.FETCHDEVICEDETAILS;
+            const requestMap = {
+              method: method,
+              params: response,
+            };
+            cy.sendMessagetoPlatforms(requestMap);
+          }
+        }
+      );
+    }
```
neeradanelxsi, line 72: where is the flag from config module which would skip fetching data for config module dynamically

neeradanelxsi: wouldn't this fail in other platforms

cypress/support/step_definitions/testSetup.js
anjalimukundan: updated as a list

cypress/support/step_definitions/testSetup.js
anjalimukundan: added an env value in config.json to determine whether dynamic value fetch is supported

cypress/support/step_definitions/testSetup.js
```diff
@@ -55,6 +55,23 @@ Given('the environment has been set up for {string} tests', (test) => {
...
+              const method = CONSTANTS.REQUEST_OVERRIDE_CALLS.FETCHDEVICEDETAILS;
+              const requestMap = {
+                method: method,
+                params: response,
+              };
+              cy.sendMessagetoPlatforms(requestMap);
+            }
+          }
+        );
+      }
```
neeradanelxsi, line 73: can we have comments here explaining about each if else here ?  
what if a config module has Cypress.env(CONSTANTS.FETCH_DEVICE_DETAILS_DYNAMICALLY_FLAG)) == true and the testype == device and there is no override for FETCHDEVICEDETAILS ? 


Did we handle that ?  Also this should be inside try /catch block and proper error  thrown

neeradanelxsi: also, what if fetching device data dynamically is a failure? how is that handled? The ticket description has that point mentioned, 

If some error occurs while fetching the source of truth, the fetching status should be updated as FAIL. Then during validation, we would need to add appropriate log mentioning that source of truth fetching failed. Hence using the data key as is.

cypress/support/step_definitions/validations.js
```diff
@@ -64,6 +64,17 @@ Given(
...
+              const type = contentObject.data[i].validations[0].type;
+              if (Cypress.env(CONSTANTS.DEVICE_DATA).hasOwnProperty(type)) {
+                contentObject.data[i].validations[0].type = Cypress.env(CONSTANTS.DEVICE_DATA)[
+                  type
+                ];
+              }
+            }
+          }
+
           // If the app ID is not passed from the feature, the default app ID will be retrieved.
```
neeradanelxsi, line 78: This would not work for localization. Only few Api's are having validation,type == 'deviceContentValidation'. 

Ideally, we should define the structure in this manner. 
if (source of truth fetch status == true && (CONSTANTS.DYNAMIC_DEVICE_DETAILS_MODULES. includes(Cypress.env(CONSTANTS.TEST_TYPE))) {) {
    search for env with the name as we use for validation and use that for validation 
}

cypress/support/step_definitions/testSetup.js
anjalimukundan: Error handling is done in configModule and configHelper module where all the cases, if token fetching, data fetching anything fails, error is thrown with proper messages

cypress/support/step_definitions/validations.js
anjalimukundan: removed deviceContentValidation check

cypress/support/step_definitions/testSetup.js
```diff
@@ -55,6 +55,21 @@ Given('the environment has been set up for {string} tests', (test) => {
...
+            const method = CONSTANTS.REQUEST_OVERRIDE_CALLS.FETCHDEVICEDETAILS;
+            const requestMap = {
+              method: method,
+              params: response,
+            };
+            cy.sendMessagetoPlatforms(requestMap);
+          }
+        }
+      );
+    }
```
neeradanelxsi, line 72: @kevinshahfws  , I feel we should remove the dependency of fetching device from FCS . Instead we can just have something like below

// If testtype == module which is included in DYNAMIC_DEVICE_DETAILS_MODULES list in constant , fetch device details dynamically
`if (CONSTANTS.DYNAMIC_DEVICE_DETAILS_MODULES.includes(Cypress.env(CONSTANTS.TEST_TYPE))) {
       // creating a request map to fetch device details from config module. Config module should have fetchDeviceDetails function inside fcs.js defined which would generate source of truth and save data in env
            const method = CONSTANTS.REQUEST_OVERRIDE_CALLS.FETCHDEVICEDETAILS;
            const requestMap = {
              method: method,
              params: response,
            };
            cy.sendMessagetoPlatforms(requestMap);
          }
        }
      );
    }`

cypress/support/step_definitions/validations.js
```diff
@@ -64,6 +64,31 @@ Given(
           const expectingError = item.expectingError;
           const isNullCase = item.isNullCase || false;
```
neeradanelxsi, line 66: add comments for each if condition here

cypress/support/constants/constants.js
```diff
@@ -99,6 +101,7 @@ module.exports = {
   DEVICE_PLATFORM: 'device.platform',
   DEVICE_VIDEORESOLUTION: 'device.videoResolution',
   DISCOVERY_LAUNCH: 'discovery.launch',
+  DYNAMIC_DEVICE_DETAILS_MODULES: ['Device'],
   EMAIL: 'email',
```
neeradanelxsi, line 105: is it possible to override this list from config module ?

cypress/support/step_definitions/validations.md
```diff
@@ -446,6 +446,25 @@ While validating, if a key is present in both fcs-validation jsons (eg: cypress/
...
+#### Format
+
+```
+    if (activeDevicedata.id == deviceId) {
+      extractedData.DEVICEID = activeDevicedata.id;
+      extractedData.DEVICE_TYPE = activeDevicedata.data.deviceType;
+    }
+    Cypress.env(CONSTANTS.DEVICE_DATA, extractedData);
+```
+
```
neeradanelxsi, line 467: For few modules, we need to generate source of truth for validating api response content. As source of truth generation varies from platform to platform, FCS is expecting the config module to generate source of truth. 

Pre-requisites - 



, the source of truth is fetched from an external API, which is dependent on the configuration module or platform on which we are testing. For the validation part, an override module called "fetchDeviceDetails" is expected to be in configModule to fetch the device details with the corresponding Urls mentioned for each config to fetch the data. DeviceId is passed from fcs to this override module as the only parameter, which is later used for fetching device details in the "configHelper" module.

The fetched data from configHelper is returned to corresponding configModule where the required data is extracted and saved to an env called "DEVICE_DATA". The data from configHelper is saved to this env as an object with the key names as the validation-key.
Initially, it maps to required object with the key as "deviceMac", if the device data is having multiple objects.
Later, for eg:, if deviceId and deviceType values are to be extracted, we will save the value into "DEVICEID" and "DEVICE_TYPE" keynames, as these are the keynames used in the validationObject.
For eg:, if activeDevicedata is the device response, the data extraction will be as follows :

cypress/support/step_definitions/validations.js
anjalimukundan: added comments

cypress/support/step_definitions/validations.md
anjalimukundan: added example data

README.md
```diff
@@ -462,3 +462,26 @@ If you want to use simplePubSub server as the means of communication for 3rd par
 7. When running FCS include env variables:
     - deviceMac: `<DEVICE_UNDER_TEST>`
     - pubSubUrl: `ws://<YOUR_IP>:8080`
+
+# Dynamic Content Validation
+
```
neeradanelxsi, line 467: this is not the documentation I was talking about. You need to add fcs.fetchDeviceDetails to this list - https://github.com/rdkcentral/firebolt-certification-suite?tab=readme-ov-file#request-overrides

cypress/support/step_definitions/validations.md
```diff
@@ -445,6 +445,31 @@ While validating, if a key is present in both fcs-validation jsons (eg: cypress/
         ]
 }
 
+# Dynamic Content Validation
```
neeradanelxsi, line 448: This is not understandable to a user how need to write their own validation. 

Currently, for validating device specific values, the validation object will have type "deviceContentValidation" . FCS will support this in two ways

- static data - explain about mac.json and where to add it
- dynamic data - explain about 

Readme should be written in terms of usage and how a user can use this feature. 

Background : 
Usage : 
Pre-requisites : 
Setup /configurations : 
How to test : 

You can keep a sepratre readme file and provide the link to that readme file here

README.md
anjalimukundan: added new override function fetchDeviceDetails in requestOverrides

cypress/support/step_definitions/validations.md
anjalimukundan: Updated the validations.md and also readme with override function

neeradanelxsi (approved):

README.md
```diff
@@ -398,6 +398,40 @@ HTTP call to the platform:<br/>
   }
   ```
 
+### fetchDeviceDetails:
```
jbigel, line 401: Is there somewhere else it would make sense for this doc to live? Just want to be careful with what we are including in the top level read me so we don't overload it with information.

cypress/support/step_definitions/validations.js
```diff
@@ -63,6 +63,32 @@ Given(
...
+                  // if the dynamic device details env contains the validation key
+                  if (Cypress.env(CONSTANTS.DEVICE_DATA).hasOwnProperty(type)) {
+                    contentObject.data[i].validations[0].type = Cypress.env(CONSTANTS.DEVICE_DATA)[
+                      type
+                    ];
+                  }
+                }
+              }
+            }
+          }
```
kevinshahfws, line 91: Do we require else condition to notify about Device data not found or fetching the data from mac.json?

README.md
neeradanelxsi: yea. we added it under the request-override heading, we can move the complete requestoverride file somewhere else and provide a link here

cypress/support/step_definitions/validations.js
anjalimukundan: Added else condition for the case were deviceData env is empty

README.md
anjalimukundan: moved the requestOverrides to a different doc file inside Doc/  folder

jbigel (dismissed):

kevinshahfws (dismissed):

cypress/support/step_definitions/validations.md
```diff
@@ -445,6 +445,88 @@ While validating, if a key is present in both fcs-validation jsons (eg: cypress/
...
+    }
+
+```
+
+When the validation mode is "deviceContentValidation", the source of truth for the validation can be taken in two ways:
+1. Static data 
+2. Dynamic data
+
+Static data :
+For static data, the source of truth is stored in a <deviceMac.json> file placed [here](../../fixtures/external/devices/) based on the deviceMac value being tested. When the test begins, these values are loaded into an object stored in an environment variable called deviceData. If deviceMac.json is not found in the specified path, the values are saved as they are, for example: DEVICEID: "DEVICEID".
```
neeradanelxsi, line 477: [here](../../fixtures/external/devices/)  link doesnt work. How will it help a user to be provided with this link ? 
 FCS does not have a external folder by default . External folder comes whenever fcs is installed with a config module. Provide the actual path of file to be added in config module repo

cypress/support/step_definitions/validations.md
```diff
@@ -445,6 +445,88 @@ While validating, if a key is present in both fcs-validation jsons (eg: cypress/
...
+
+Dynamic data :
+For dynamic data, the source of truth is fetched from the configModule's override function called "fetchDeviceDetails" explained [here](https://github.com/rdkcentral/firebolt-certification-suite?tab=readme-ov-file#request-overrides).
+
```
neeradanelxsi, line 481: Usage has to be properly explained.

cypress/support/step_definitions/validations.md
```diff
@@ -445,6 +445,88 @@ While validating, if a key is present in both fcs-validation jsons (eg: cypress/
...
+If the dynamic device details fetching fails in configModule, the data will default to the static data. FCS expects the source of truth to be generated for the modules listed in [DYNAMIC_DEVICE_DETAILS_MODULES](../constants/constants.js). 
+
+
+### Implementation :
+
+The configModule should have an override function "fetchDeviceDetails" as explained [here](https://github.com/rdkcentral/firebolt-certification-suite?tab=readme-ov-file#request-overrides). This function should fetch the data and return it in the format specified below. If deviceId, deviceType, and distributor values need to be extracted, they should be saved as "DEVICEID", "DEVICE_TYPE", and "DEVICE_DISTRIBUTOR" keys, as these are the key names used in the validationObject. These values will then replace the values in <device-mac.json> stored in the environment variable "DEVICE_DATA" in FCS.
+
```
neeradanelxsi, line 492: lines 493 to 494 is comcast specific details or rather config module specific details. Why to add it here

neeradanelxsi: is this how sky fetch data ?

cypress/support/step_definitions/validations.md
```diff
@@ -445,6 +445,88 @@ While validating, if a key is present in both fcs-validation jsons (eg: cypress/
...
+The response by overriding the static values in the environment will be as follows:
+Example:
+
+  ```
+  {
+    "DEVICEID": "354444327",
+    "DEVICE_TYPE": "ipstb",
+    "DEVICE_MODEL": "VALUE",
+     ...
```
neeradanelxsi, line 525: Use case : User want to add DEVICE_DISTRIBUTOR value to dynamic source of truth fetching. How can he add that ? Where is the explanation for that starting from writing a validation object with type : "DEVICE_DISTRIBUTOR" to returning the device_data containing the key DEVICE_DISTRIBUTOR ?

neeradanelxsi: create a new reamde file and provide link from here

cypress/support/step_definitions/validations.md
anjalimukundan: added a generic path which is common for all configModule

cypress/support/step_definitions/validations.md
anjalimukundan: updated

cypress/support/step_definitions/validations.md
anjalimukundan: updated

cypress/support/step_definitions/validations.md
anjalimukundan: removed platform specific data

Abhishk123 (approved):

PR #191, opened 2024-08-08
second phase of semantic PR

jbigel:


.github/workflows/pr-name-check.yaml
```diff
@@ -0,0 +1,28 @@
...
+    branches:
+      - 'dev'
+    types: [opened, edited, synchronize]
+
+jobs:
+  check-pr-name:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@v2
```
ksentak, line 14: Can we change this to v4? I believe v2 uses a deprecated version of node which posts warnings that will soon become errors in GitHub.

ksentak (changes requested):
Small comment

ksentak (approved):

PR #192, opened 2024-08-09
fix: undefined validation object fix

Nandana-NNR:


ibrahimshuail (approved):

cypress/support/step_definitions/validations.js
```diff
@@ -177,30 +177,37 @@ Then(/'(.+)' will (be|stay) in '(.+)' state/, (app, condition, state) => {
...
+          );
+        }
       }
     }
-  } else {
+  } catch (error) {
+    throw new Error(
+      `Following error occurred during custom validation: ${JSON.stringify(error.message)}`
+    );
+  }
```
neeradanelxsi, line 209: i think we should only print error here, right? suppose if you get an error and there is no message field?

neeradanelxsi, line 209: also, is there an else condition for line 185 ? What if there is no validationObject ?

cypress/support/step_definitions/validations.js
```diff
@@ -177,30 +177,37 @@ Then(/'(.+)' will (be|stay) in '(.+)' state/, (app, condition, state) => {
...
+  if (!validationObject) {
```
neeradanelxsi, line 210: not really understanding the purpose of moving else condition outside to line 210. Cant we do 
```

try {
      if (validationObject) {
        changes from line 186 to 204
      }else{
         changes from lines 211 to 216
       }
}catch(error) {
     Here, give a generic error message 
     throw new Error(
      `Error occurred during validation: ${JSON.stringify(error)}`
    );
}
```

jbigel (dismissed):

cypress/support/step_definitions/validations.js
Nandana-NNR: Addressed

cypress/support/step_definitions/validations.js
```diff
@@ -177,36 +177,40 @@ Then(/'(.+)' will (be|stay) in '(.+)' state/, (app, condition, state) => {
...
-    cy.validateLifecycleState(appObject.getAppObjectState().state, appId);
-    cy.validateLifecycleHistoryAndEvents(
-      appObject.getAppObjectState().state,
-      appId,
-      isEventsExpected
-    );
+  } catch (error) {
+    throw new Error(`Following error occurred during custom validation: ${JSON.stringify(error)}`);
   }
 });
```
neeradanelxsi, line 215: throw new Error(`Following error occurred during custom validation: ${JSON.stringify(error)}`); to be changed to 
throw new Error(`Error occurred during validation: ${JSON.stringify(error)}`);

neeradanelxsi (approved):
Minor comment on language for log

neeradanelxsi (approved):

PR #195, opened 2024-08-16
fix: Exempted methods does not have error null check for error

Abhishk123:


cypress/support/cypress-commands/assertion.js
```diff
@@ -175,8 +175,15 @@ Cypress.Commands.add(
...
-            if (!UTILS.getEnvVariable(CONSTANTS.IS_SCENARIO_EXEMPTED, false)) {
+            // Skip the null check for exempted scenarios based on the error response.
+            if (UTILS.getEnvVariable(CONSTANTS.IS_SCENARIO_EXEMPTED, false)) {
+              // Performing error null check validation if the response error is not null.
+              if (response?.error == null) {
+                cy.errorNullCheck(response, CONSTANTS.ERROR).then((result) => {
+                  validationCheck.push(result);
+                });
+              }
+            } else {
```
neeradanelxsi, line 186: why we need to skip this check if it's an error?
Can't we do similar to negative cases where we are expecting error ? 

[8/18/2024 6:45:05 AM][loggingValidationCheckResult][Error null Check: PASS. Expected error in response: "entityId" is a required property ]

Abhishk123: addressed the comment by adding this line

neeradanelxsi (approved):

jbigel (approved):

PR #198, opened 2024-08-20
Feat: Player interaction logs validation

Eswar2103:


kevinshahfws (approved):

kevinshahfws (approved):

cypress/support/cypress-support/src/main.js
```diff
@@ -126,8 +126,14 @@ export default function (module) {
 
   // beforeEach
   beforeEach(() => {
+    UTILS.getEnvVariable(CONSTANTS.FB_INTERACTIONLOGS).clearLogs();
     cy.getBeforeOperationObject();
     UTILS.destroyGlobalObjects([CONSTANTS.LIFECYCLE_APP_OBJECT_LIST]);
+    let scenarioName = cy.state().test.title;
+    if (scenarioName.includes('(example')) {
```
jbigel, line 133: Reading the code, it is unclear why we would need to modify the scenario name when `example` exists. Can we add comments to make it clear.

Eswar2103: Removed the above code and using it from before operation object

jbigel (approved):

PR #202, opened 2024-08-25
feat: Changed suite default communication mode to Transport

Eswar2103:


cypress/support/cypress-support/src/main.js
```diff
@@ -321,19 +321,19 @@ export default function (module) {
...
       (!additionalParams[CONSTANTS.COMMUNICATION_MODE] ||
         additionalParams[CONSTANTS.COMMUNICATION_MODE]) &&
-      UTILS.getEnvVariable(CONSTANTS.COMMUNICATION_MODE, false)
+      UTILS.getEnvVariable(CONSTANTS.SUITE_COMMUNICATION_MODE, false)
     ) {
       additionalParams[CONSTANTS.COMMUNICATION_MODE] = UTILS.getEnvVariable(
-        CONSTANTS.COMMUNICATION_MODE
+        CONSTANTS.SUITE_COMMUNICATION_MODE
       );
     }
```
neeradanelxsi, line 338: do we really need to create a new constant here? The only requirement is to keep sanity default execution as Transport, if someone need to execute via SDK, they can just change the communicationMode to SDK by passing in CLI right? 
If we keep a new constant, we need that to be passed right

Eswar2103: I added a new environment variable because if there is only one communication variable for both modules and suites, it will be defaulted to *SDK*, unless someone changes it to *Transport* for suites. It will execute with SDK mode only.

So if you want to keep one communicatioMode variable it can be done too, however in CI both SDK and Transport are executed, but if someone used FCS then the default will be SDK unless they override it.

jbigel (approved):

kevinshahfws (approved):

PR #205, opened 2024-08-27
feat: Add support for skipping individual scenarios from beforeEach operation

jmeier204:


neeradanelxsi (approved):

cypress/support/cypress-commands/commands.js
```diff
@@ -575,6 +575,11 @@ Cypress.Commands.add('getBeforeOperationObject', () => {
     beforeOperation = scenarioList[scenarioName].beforeOperation;
     if (Array.isArray(beforeOperation)) {
       cy.get(Object.values(beforeOperation)).each((beforeOperationObject) => {
+        if (beforeOperationObject.hasOwnProperty('skipTest')) {
+          if (beforeOperationObject.skipTest === true) {
+            UTILS.skipCurrentTest();
+          }
+        }
```
ksentak, line 582: I don't think you need the double if statement here. You should be able to just do:
```
if (beforeOperationObject.skipTest) {
  UTILS.skipCurrentTest();
}
```

Reason being is because `beforeOperationObject.skipTest` will evaluate to false if it is undefined, null, or false, so there's no need to check if the property exists and if it's true.

jbigel (approved):

neeradanelxsi (approved):

PR #243, opened 2024-10-28
fix: Fixing clear event listener issue

Abhishk123:


cypress/fixtures/fireboltMocks/clearEventData.json
```diff
@@ -42,5 +42,9 @@
     "SECONDSCREEN_ONCLOSEREQUEST_EVENT":{
         "event": "secondscreen.onCloseRequest",
         "firstParty": false
+    },
+    "ADVERTISING_ONSKIPRESTRICTIONCHANGED_EVENT":{
+        "event": "manage_advertising.onSkipRestrictionChanged",
+        "firstParty": false
     }
```
neeradanelxsi, line 49: why is firstParty false ? this looks to be a manage event . 

is this for clearing event listener on FCA , but for a manage event ?

neeradanelxsi: if this is applicable only for xumo config module, cant we place this in xumo module and not directly in fcs >?

jbigel (approved):

cypress/fixtures/fireboltMocks/clearEventData.json
Abhishk123: move this to config module and firstParty is false because making a call to 3rd party app

neeradanelxsi (approved):

PR #268, opened 2024-11-20
feat: Deep link implementation

Abhishk123:


cypress/plugins/common.js
```diff
@@ -18,7 +18,7 @@
 const fs = require('fs');
 const testDataProcessor = require('./testDataProcessor');
 const logger = require('../support/Logger')('common.js');
-const { generateIndexFile, preprocessDeviceData } = require('./pluginUtils');
+const { generateIndexFile, preprocessDeviceData, fetchAppMetaData } = require('./pluginUtils');
 const CONSTANTS = require('../support/constants/constants');
```
neeradanelxsi, line 22: what is this used for ? after importing ?

cypress/plugins/common.js
```diff
@@ -46,6 +46,8 @@ function genericSupport(config) {
   generateIndexFile(CONSTANTS.FIREBOLTMOCK_FROM_FCS, 'fireboltMocks');
   generateIndexFile(CONSTANTS.FIREBOLTMOCK_FROM_CONFIGMODULE, 'fireboltMocks');
 
+  generateIndexFile('cypress/fixtures/external/intentTemplates/', 'intentTemplates');
+
   // The sequence of override - the default config in the config.js file, overriden by supportConfig.json and then by the command line arguments.
```
neeradanelxsi, line 51: also, what is line 49 doing ?

cypress/support/cypress-commands/commands.js
```diff
@@ -1623,3 +1669,67 @@ Cypress.Commands.add('initiatePerformanceMetrics', () => {
...
 });
+
+/**
+ * @module commands
+ * @function fetchAppMetaData
+ * @descriptionReads app metadata from the appData directories of both fcs and configModule, then combines them, prioritizing the configModule metadata.
+ * @example
+ * cy.fetchAppMetaData()
+ */
+Cypress.Commands.add('fetchAppMetaData', () => {
```
neeradanelxsi, line 1680: cant these path be moved to constants

cypress/support/step_definitions/testSetup.js
```diff
@@ -106,6 +109,14 @@ Given('the environment has been set up for {string} tests', (test) => {
...
     }
   }
+
+  if (test === 'Deeplink' && !Cypress.env('intentTemplates') && !Cypress.env('app_metadata')) {
+    cy.fetchAppMetaData().then((appMetaData) => {
+      Cypress.env('app_metadata', appMetaData);
+    });
+    const combinedIntentTemplates = _.merge(internalIntentTemplates, externalIntentTemplates);
+    Cypress.env('intentTemplates', combinedIntentTemplates);
+  }
```
neeradanelxsi, line 119: This cannot go in FCS. it has to be in config module I believe. Can be confirmed in calls

cypress/support/cypress-commands/commands.js
```diff
@@ -883,114 +884,159 @@ Cypress.Commands.add('launchApp', (appType, appCallSign, deviceIdentifier) => {
...
+      // Check if intentTemplate exists for native appType
+      else if (
+        appType === 'native' &&
+        intentTemplates[appType][appId] &&
+        intentTemplates[appType][appId][intent]
+      ) {
+        intentTemplate = intentTemplates[appType][appId][intent];
+      }
+      // Log failure if intentTemplate is not found
+      else {
```
neeradanelxsi, line 924: is it acceptable to keep native here ?

Abhishk123: moved to config

cypress/support/step_definitions/testSetup.js
Abhishk123: moved to config module

cypress/support/cypress-commands/commands.js
Abhishk123: yes, moved to constants

cypress/support/cypress-support/src/utils.js
```diff
@@ -1001,20 +1001,46 @@ global.resolveAtRuntime = function (input) {
...
         }
 
-        // If a function name is present in the pattern, call the function with pattern content as input.
-        // Reading the pattern content from the runtime environment variable
-        const value = runtimeEnv[pattern] !== undefined ? runtimeEnv[pattern] : match;
-        return functionName && functions[functionName] ? functions[functionName](value) : value;
+        // If the input contains '->', Consdering it as an object path and extracting the value from the object
+        if (pattern.includes('->')) {
+          return extractValueFromObjectPath(pattern);
+        } else {
```
jbigel, line 1007: Here I think instead of an 'else' we can just check if there is a functionName and it should handle the case I mentioned in the call this morning.

jbigel (commented):
Code looks good for the most part outside of the one change for `resolveAtRuntime`. For the documentation I'm wondering if we need more details on what intent is and how to use it as well as app data.

cypress/support/cypress-support/src/utils.js
Abhishk123: addressed the comment
First we are going to fetch the value from the object then calling function to change the case

jbigel (approved):

ibrahimshuail (approved):

PR #271, opened 2024-11-21
BREAKING CHANGE: Update "launchApp" intent addon to add additional params

Eswar2103:


cypress/support/cypress-support/src/utils.js
```diff
@@ -1162,6 +1162,35 @@ class InteractionsLogs {
 const interactionLogs = new InteractionsLogs();
 Cypress.env(CONSTANTS.FB_INTERACTIONLOGS, interactionLogs);
 
+/**
+ * @module utils
+ * @function censorPubSubToken
+ * @description A Function to sensor the pubSubToken from the launch intent.
+ * @param {String} data - The intent to be parsed and censored.
+ * @example
+ * censorPubSubToken('{"method: "launch", "params": {"intent": {"data": {"query": '{"params": {"pubSubToken": "ey3456789ab}'}}}}')
```
mhughesacn, line 1171: Please confirm that the "pubSubToken" value is not sensitive.

Eswar2103: It is a sensitive value, that's why it is being censored and it is not being stored anywhere, just masking it when showing in report, but the original value can be passed on. 
This pubSubToken is not getting generated in this code, it comes from external source and go to another services, this is just a pass through.

mhughesacn: Sorry @Eswar2103, perhaps I wasn't clear: Is there any issue with the exact magic string "ey....ab" being publicly visible in the source code of this opensourced repository?

Eswar2103: If you are asking about the description added, I have corrected it now to some generic value.

Eswar2103: @mhughesacn it is a dummy value.

jbigel (approved):

kevinshahfws (approved):

PR #279, opened 2024-11-28
feat: [FCS] Implement callPlayerMethods()

Abhishk123:


cypress/plugins/pluginUtils.js
```diff
@@ -155,4 +156,40 @@ function preprocessDeviceData(config) {
   }
 }
 
-module.exports = { getAndDereferenceOpenRpc, generateIndexFile, preprocessDeviceData };
+function fetchPlayerMethodObject(config) {
+  const fcsPlayerMethods = 'cypress/fixtures/objects/playerMethodObject.json';
+  const configModulePlayerMethods = 'cypress/fixtures/external/objects/playerMethodObject.json';
```
neeradanelxsi, line 161: i dont think we can keep player these in fcs as IPA is not open source

PreethiMaai: addressed

cypress/plugins/pluginUtils.js
```diff
@@ -155,4 +156,55 @@ function preprocessDeviceData(config) {
...
 }
 
-module.exports = { getAndDereferenceOpenRpc, generateIndexFile, preprocessDeviceData };
+/**
+ * @function fetchWrapperMethodObject
+ * @description Loads and combines two `wrapperMethodObject.js` files from FCS and config, then assigns the combined object to env.
+ * @param {object} config - The config object.
+ * @example
+ * fetchWrapperMethodObject(config);
+ */
```
neeradanelxsi, line 165: please do confirm if this can be kept in FCS

cypress/plugins/pluginUtils.js
```diff
@@ -155,4 +156,55 @@ function preprocessDeviceData(config) {
...
+function fetchWrapperMethodObject(config) {
+  const fcsWrapperMethodObjectPath = path.resolve(
+    __dirname,
+    '..',
```
neeradanelxsi, line 169: what if a platform /config module does not have this js file ?

is it really necessary to get this js file here ? 

for dismiss and deeplink, we are not processing it here, rather in env setup step only. 

as this is only applicable to IPA, cant we follow the same here

anjalimukundan:
closed without merging

PR #283, opened 2024-12-05
Log Correction

ibrahimshuail:
Corrected the fire log for "Original Response to be converted to firebolt equivalent"


ibrahimshuail:
closing the PR

PR #289, opened 2024-12-09
feat: Incorporated FCS setter into sendMessageToplatform

kdivya153:


cypress/support/cypress-support/src/globalSetter.js
```diff
@@ -0,0 +1,31 @@
...
+
+global.setterFailure = (message, error) => {
+  const methodName = UTILS.getEnvVariable(CONSTANTS.REQUEST_OVERRIDE_METHOD);
+  const errorMessage = `${methodName} ${message || `FCS setter for fcsSetter.${methodName} failed`}`;
+
+  // cy.then() added to handle the errors gracefully
+  cy.then(() => {
+    const thrownError = new Error(errorMessage);
+    if (error) {
+      thrownError.message += `error due to : ${JSON.stringify(error)}`;
```
kschrief, line 18: If we set "thrownError.cause" to the original error we can have the "cause" of our error be the error that was given to us.

cypress/support/cypress-support/src/globalSetter.js
```diff
@@ -0,0 +1,31 @@
...
+
+global.setterSuccess = (message = 'successful response') => {
+  const methodName = UTILS.getEnvVariable(CONSTANTS.REQUEST_OVERRIDE_METHOD);
+  cy.log(`[${methodName}] Success: ${message}`);
+  return { status: 'success', message };
+};
+
+global.setterFailure = (message, error) => {
+  const methodName = UTILS.getEnvVariable(CONSTANTS.REQUEST_OVERRIDE_METHOD);
+  const errorMessage = `${methodName} ${message || `FCS setter for fcsSetter.${methodName} failed`}`;
```
kschrief, line 12: It looks like the formatting is different than for success? Let's have the messaging be consistent:

"[Setter Method (function)] (message)"

Ex: "[Setter Method fcs.setClosedCaptions] FCS Setter failed/succeeded"

cypress/support/cypress-support/src/globalSetter.js
```diff
@@ -0,0 +1,31 @@
...
+      thrownError.message += `error due to : ${JSON.stringify(error)}`;
+    }
+
+    throw thrownError;
+  });
+};
+
+global.setterNotImplemented = (message) => {
+  const methodName = UTILS.getEnvVariable(CONSTANTS.REQUEST_OVERRIDE_METHOD);
+  const errorMessage = `${message || `FCS setter for fcsSetter.${methodName} does not have an implementation`}`;
```
kschrief, line 27: For now let's have this be the same format as success/failure. We can add additional logs in the next ticket when we link to the documentation.

cypress/support/cypress-support/src/main.js
```diff
@@ -253,14 +254,36 @@ export default function (module) {
...
-        const message = await config.getRequestOverride(requestMap);
-        // perform MTC call/FB call only if the message is not null
-        if (message != null) {
-          const response = await transport.sendMessage(message);
-          const result = config.getResponseOverride(response);
-          resolve(result);
+        const [moduleName, methodName] = requestMap.method.split('.');
+        Cypress.env(CONSTANTS.REQUEST_OVERRIDE_METHOD, methodName);
+        // Check if request is for fcs setters
+        if (moduleName === 'fcsSetter') {
```
kschrief, line 260: Let's add "fcsSetter" as a constant as well

cypress/support/cypress-support/src/main.js
```diff
@@ -253,14 +254,36 @@ export default function (module) {
...
+          const method = fcsSetter[methodName];
+          if (method && typeof method === 'function') {
+            try {
+              const { attribute, value } = requestMap.params;
+              const response = await method(attribute, value);
+              resolve(response);
+            } catch (error) {
+              resolve(
```
kschrief, line 268: We don't want to resolve here. Whatever error we get here (Probably from a setterFailure or setterNotImplemented) should bubble up and fail the step.

Calling setterFailure() again here is redundant, especially if we got here because of a setterNotImplemented()

cypress/support/cypress-support/src/main.js
```diff
@@ -253,14 +254,36 @@ export default function (module) {
...
+                setterFailure(`Error while invoking ${requestMap.method} due to ${error.message}`)
+              );
+            }
+          } else {
+            resolve(setterNotImplemented(`Method ${methodName} not implemented`));
```
kschrief, line 273: We don't want to resolve here either. We can reject if we need to do something but we shouldn't be resolving

cypress/support/e2e.js
```diff
@@ -17,6 +17,7 @@
  */
 
 // Importing generic module
```
kschrief, line 19: This comment should either go away or stay above "import * as genericModule"

cypress/support/cypress-support/src/fcsSetter.js
```diff
@@ -0,0 +1,10 @@
+const setClosedCaptions = async (attribute, value) => {
```
kschrief, line 1: This file shouldn't be here.

We still want "fcsSetters" to be in the request overrides of the config module.

Also the file/module should be "fcsSetters" with an s on the end.

kschrief (changes requested):

cypress/support/cypress-support/src/globalSetter.js
kdivya153: Addressed

cypress/support/cypress-support/src/globalSetter.js
kdivya153: Addressed

cypress/support/cypress-support/src/main.js
kdivya153: Added in the constants

cypress/support/cypress-support/src/main.js
kdivya153: Addressed

cypress/support/e2e.js
kdivya153: Done

cypress/support/cypress-support/src/fcsSetter.js
kdivya153: Created an helper function to get from config module dynamically

cypress/support/cypress-support/src/config.js
```diff
@@ -42,6 +42,17 @@ export default class Config {
     return this.moduleName;
   }
 
+  /* get fcsSetterOverride methods from configuration module.
+   * ex: {"method": "fcsSetters.setClosedCaptions", "params": {"attribute":enable "value": true }}
+   * If method not present, return null
+   * If method present, return the object to caller.
+   */
+
+  getFcsSetterOverride(moduleName, methodName) {
```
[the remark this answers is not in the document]
kdivya153: Renamed to getRequestOverride()

kschrief (dismissed):

cypress/support/cypress-support/src/main.js
```diff
@@ -252,15 +252,49 @@ export default function (module) {
...
-          const result = config.getResponseOverride(response);
-          resolve(result);
+      return new Promise(async (resolve, reject) => {
+        const [moduleName, methodName] = requestMap.method.split('.');
+        Cypress.env(CONSTANTS.REQUEST_OVERRIDE_METHOD, methodName);
+        // Check if request is for fcs setters
+        if (moduleName === CONSTANTS.FCS_SETTER) {
+          const method = config.getRequestOverride(moduleName, methodName);
+          if (typeof method === 'function') {
+            const params = requestMap.params || {};
```
kdivya153, line 262: incorporated logic to pass the call the fcsSetter() based on the arguments

cypress/support/cypress-support/src/config.js
```diff
@@ -42,7 +42,18 @@ export default class Config {
     return this.moduleName;
   }
 
-  /* get request override config from configuration module.
+  /* get fcsSetterOverride methods from configuration module.
```
kschrief, line 45: This can be changed to "request override"

kschrief (changes requested):

kschrief (approved):

PR #299, opened 2024-12-19
feat: Adding fcsSetter template in default config module

kdivya153:


defaultModule/transportLayer/index.js
kdivya153, line 1: In the Config Module, the directory is named transportLayer, but the default module had it as transportLayers. To prevent an empty transportLayers directory from being copied during yarn install, renamed it to match the Config Module.

Scripts/checkDefaultConfig.js
```diff
@@ -40,10 +40,59 @@ const ensureConfigModule = async () => {
...
+          const defaultContent = await fs.readFile(defaultIndexPath, 'utf8');
+          const targetContent = targetIndexExists ? await fs.readFile(targetIndexPath, 'utf8') : '';
+
+          // Regular expression to remove comments (single-line and multi-line)
+          const commentRegex = /\/\/.*|\/\*[\s\S]*?\*\//g;
+
+          // Remove comments from default content
+          const cleanedDefaultContent = defaultContent.replace(commentRegex, '');
+
+          // Merge contents, remove duplicates
```
kschrief, line 79: What happens if both the config module AND default module both have the same file being exported?

kdivya153: merging process add exports from defaultModule/index.js to configModule/index.js only if they don’t already exist, ensuring configModule content takes precedence.

kschrief (approved):

PR #306, opened 2025-01-07
feat: Validate initialize int player

anjalimukundan:


cypress/plugins/index.js
```diff
@@ -212,6 +212,15 @@ module.exports = async (on, config) => {
     checkFileExists(filePath) {
       return fs.existsSync(filePath);
     },
+    loadJSFile(filePath) {
+      console.log('filepath 2652', filePath);
```
jbigel, line 216: This looks like a debug log, can we remove it.

cypress/plugins/index.js
```diff
@@ -212,6 +212,15 @@ module.exports = async (on, config) => {
...
+      try {
+        const resolvedPath = path.resolve(__dirname, '..', filePath);
+        return require(resolvedPath);
+      } catch (error) {
+        return null;
```
jbigel, line 221: Curious about returning null instead of throwing an error? If null was a design decision could we at least print an error log?

cypress/support/cypress-commands/commands.js
```diff
@@ -1760,3 +1761,89 @@ Cypress.Commands.add('extractAppMetadata', (appDataDir, appMetaDataFile) => {
...
+    const isPresent = interactionLogs.some((log) => {
+      log.method === method;
+    });
+    if (!isPresent) {
+      methodNotFound.push(method);
+    }
+  });
+  if (methodNotFound.length > 0) {
+    const message = `The following methods are missing in interactionLogs: [${methodNotFound}].`;
+    // const message = `Methods in [ ${methodNotFound} ] are not present in interactionLogs`;
```
jbigel, line 1804: can we remove commented code?

anjalimukundan: removed the commented line

PR #307, opened 2025-01-08
feat: test case versioning changes

PreethiMaai:


Scripts/copyExternalConfigData.js
```diff
@@ -19,31 +19,11 @@ const fs = require('fs');
 const path = require('path');
 const logger = require('../cypress/support/Logger')('copyExternalConfigData.js');
 
-logger.info(
-  'Copying Config fixtures into fixtures/external and Testcases into TestCases/Distributor'
-);
+logger.info('Copying Config fixtures and Testcases into sdkResources/external');
 
-// Config for testCase
-const EXTERNAL_DIR_TESTCASE = path.join(__dirname, '..', 'cypress', 'TestCases', 'Distributor');
```
neeradanelxsi, line 27: curious on why we removed 'node_modules', 'configModule', etc from 28 and 39. will that cause any issues

sdkResources/0.8.1/TestCases/sample.feature
```diff
@@ -0,0 +1,6 @@
+# sample feature file for testing. Will remove this before merging
```
jbigel, line 1: Do we want to merge this to dev?

specHelperConfig.js
```diff
@@ -28,14 +28,14 @@ const logger = require('./cypress/support/Logger')('specHelperConfig.js');
 function getSpecPattern(testSuite) {
   const patterns = {
     module: [
-      '**/FireboltCertification/*.feature',
-      '**/FireboltCertification/*/*.feature',
-      '**/Distributor/*.feature',
-      '**/Distributor/*/*.feature',
+      'cypress/TestCases/FireboltCertification/*.feature',
```
jbigel, line 31: Please correct me if I am wrong but I don't think any of these changes are necessary since the test cases path hasn't actually changed at the time cypress is looking for the files?

PreethiMaai: Earlier, we had feature files inside `cypress/TestCases`, and the` **/ pattern` was used to search through all folders and subfolders, which didn’t cause any issues. However, now `cypress/TestCases` contains specific test cases, while the sdkResources folder holds test cases for all versions. If we use the old spec pattern, it would fetch test cases from both cypress and sdkResources. So, I updated the pattern to target `cypress/TestCases `specifically.
image

Scripts/copyFixturesTestcases.js
```diff
@@ -0,0 +1,86 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+const fs = require('fs');
+const path = require('path');
+
+const sdkVersion = process.env.sdkVersion || 'latest';
```
jbigel, line 21: `latest` is no longe a valid or possible scenario at this point of the code right? Can we update here?

Scripts/run.js
```diff
@@ -1,10 +1,22 @@
...
 
 // Reading first parameter from the scripts to call function
 const functionName = process.argv[2];
 const params = process.argv.slice(3).join(' ');
 
+const sdkVersionMatch = params.match(/sdkVersion=([^\s,]+)/);
+let sdkVersion = sdkVersionMatch ? sdkVersionMatch[1] : 'latest';
+
+// Access SDK_VERSION_LATEST from the config
+const sdkVersionLatest = packageJson.config.SDK_VERSION_LATEST;
```
jbigel, line 15: We really only need to read this variable and object if `sdkVersion === 'latest'`. We can combine this line with line 18.

specHelperConfig.js
jbigel: Okay that makes sense. Thank you for clarifying.

Scripts/run.js
PreethiMaai: updated new logic

README.md
```diff
@@ -94,6 +94,27 @@ By default, the project uses a predefined configuration module referred to as th
 
 Once complete, continue following the rest of the [Setup Instructions](#setup).
 
+#### Config Variables
+
+The Firebolt Certification Suite (FCS) uses configuration variables to manage the behavior of the suite during runtime. These variables are defined in the `package.json`.
```
jbigel, line 99: Can we emphasize how this differs from the other env variables in cypress.config.js and is the only configuration variable located in package.json.

README.md
```diff
@@ -94,6 +94,27 @@ By default, the project uses a predefined configuration module referred to as th
...
+
+##### SDK Version
+- **Variable Name**: `supportedSDKVersion `
+- **Purpose**: Specifies the SDK version that FCS should use during tests.
+- **Default Value**: `"1.4.1"`
```
jbigel, line 104: This will change whenever 1.4.1 is not the latest.

cypress/TestCases/README.md
```diff
@@ -1,33 +1,20 @@
-# Firebolt Certification
+# Test Cases
 
-Feature files for Firebolt Certification Modules.
+This directory contains the test cases that are used during the execution of the Firebolt Certification Suite (FCS).
```
jbigel, line 3: Can you emphasize that it is empty in the repo because it is cleared and copied over at the beginning of each run.

jbigel: It looks like you do explain for the most part but we should put this info at the top as well because people familiar with cypress will be confused.

cypress/fixtures/README.md
```diff
@@ -1,34 +1,20 @@
...
-- versions: Contains different versions of firebolt.json includes versioning, capabilities and openrpc.
-
-## Firebolt Calls and Validation support:
-
-1. [Static JSON Objects](docs/staticJSONObjects.md)
-2. [Static JS Objects](docs/staticJSObjects.md)
-3. [Dynamic JS Objects](docs/dynamicObjects.md)
+# Fixtures
+
+This directory contains the fixture files that are used during the execution of the Firebolt Certification Suite (FCS).
```
jbigel, line 3: same as above.

cypress/TestCases/README.md
```diff
@@ -1,33 +1,20 @@
-# Firebolt Certification
+# Test Cases
```
jbigel, line 1: When there are other files in the TestCases folder does this appear at the top? There should be a way to force that to happen.

README.md
PreethiMaai: Updated the README to mention that this is the only config variable in package.json, unlike the others in cypress.config.js.

README.md
PreethiMaai: Removed the specific version (1.4.1) and updated it to a more generic statement.

cypress/TestCases/README.md
PreethiMaai: I’ve updated the logic to ensure that the README in cypress/fixtures and TestCases is not removed, so it will always appear at the top.

neeradanelxsi (approved):

jbigel (dismissed):

ibrahimshuail (approved):

PR #315, opened 2025-01-17
feat: Added Implementation for FcsSetter

kdivya153:


Scripts/checkDefaultConfig.js
```diff
@@ -65,30 +65,50 @@ const ensureConfigModule = async () => {
...
+          );
 
-          // Write back to target index.js
-          await fs.writeFile(targetIndexPath, mergedContent, 'utf8');
+          // Add default lines only if the variable name does not exist in the target
+          const mergedLines = [...targetLines];
+          defaultLines.forEach((line) => {
+            if (/=/.test(line)) {
+              const variableName = extractVariableName(line);
+              if (!targetKeys.has(variableName)) {
```
kdivya153, line 104: Joe brought a concern that if same variablename has different datatype then it creates 2 copies.So leverage the logic to ignore the datatype and check only for the variable name.

cypress/support/cypress-commands/commands.js
```diff
@@ -1209,6 +1209,10 @@ Cypress.Commands.add('sendMessageToPlatformOrApp', (target, requestData, task) =
       fireLog.assert(false, `Invalid ${target} target, it should be either app or platfrom`);
     }
   }).then((response) => {
+    if (method.startsWith(CONSTANTS.FCS_SETTER)) {
+      console.log('Skipping schema validation for FCS_SETTER method');
```
kschrief, line 1213: This message should show the actual "fcsSetters" instead of the name of the constant.

defaultModule/requestModules/fcsSetters_firebolt.js
```diff
@@ -0,0 +1,543 @@
...
+        throw new Error('Invalid response: Response is null or not an object');
+      }
+      //Validating the response
+      await validateResponse(requestMap, response, setterMethod);
+
+      const success = response.hasOwnProperty(CONSTANTS.RESULT);
+      if (success) {
+        return await setterSuccess(`Launched ${value} successfully!`);
+      } else {
+        return await setterFailure(`Failed to launch app ${value}`);
```
kschrief, line 27: Can we print the response object if we get a failure? This should apply to the rest of the setters as well.

defaultModule/requestModules/fcsSetters_firebolt.js
```diff
@@ -0,0 +1,543 @@
...
+
+/**
+ * Sets additional metadata for localization or configuration.
+ * @param {any} value - The value to set.
+ * @returns {Promise<any>} A promise that resolves/reject based on the response when the operation is complete.
+ * @example fcsSetters.setAdditionalInfo("{ key: 'exampleKey', value: 'exampleValue' }");
+ */
+const setAdditionalInfo = async (attribute, value) => {
+  const setterMethod = `fcsSetters.${setAdditionalInfo.name}`; // Dynamically retrieve the function name to map response
+  const requestMap = createRequestMap(`localization.set${attribute}`, value);
```
kschrief, line 44: Does this work or is there a "setAdditionalMetadat" field? Do we know what attributes will work here?

kschrief (changes requested):

defaultModule/requestModules/fcsSetters_firebolt.js
kdivya153: Addressed - Added console log for printing the response on failure for all the methods

defaultModule/requestModules/fcsSetters_firebolt.js
kdivya153: Kurt,As per excel sheet it supports two scenerio's removeAdditionalInfo and addAdditionalInfo,where we'll be sending Key and values as param.I dont see any Firebolt API method like "setAdditionalMetadata" in Localization other than add and remove.
Reference:
https://github.com/rdkcentral/firebolt-certification-suite/blob/1849bcb666c37282c6978c00cfb141984eedadef/cypress/TestCases/FireboltCertification/Manage/LocalizationManage.feature#L62
https://github.com/rdkcentral/firebolt-certification-suite/blob/1849bcb666c37282c6978c00cfb141984eedadef/cypress/fixtures/fireboltCalls/localization.js#L135

defaultModule/requestModules/fcsSetters_firebolt.js
```diff
@@ -0,0 +1,556 @@
...
+
+/**
+ * Sets the country code.
+ * @param {any} value - The value to set.
+ * @returns {Promise<any>} A promise that resolves/reject based on the response when the operation is complete.
+ * @example fcsSetters.setCountryCode("US");
+ */
+const setCountryCode = async (value) => {
+  const setterMethod = `fcsSetters.${setCountryCode.name}`; // Dynamically retrieve the function name to map response
+  const requestMap = createRequestMap(`localization.setcountryCode`, value);
```
kschrief, line 113: Per the official spec this is "setCountryCode" with a capital "Country"

defaultModule/requestModules/fcsSetters_firebolt.js
```diff
@@ -0,0 +1,556 @@
+import { apiObject } from '../../../cypress/support/appObjectConfigs.js';
```
kschrief, line 1: This file should be portable. We should remove these imports.

Also as discussed we will not be performing the "validateResponse()" logic in this file. The setterSuccess() global function will do that but with a dummy FB response. This will let the fcsSetters method do its own validation for its own platform and FCS will see the dummy response as a successful FB response that passes schema validation (AKA it's not breaking the FCS flow)

defaultModule/requestModules/fcsSetters_firebolt.js
kschrief: Per your second link we're calling "Localization.setAdditionalInfo()" but your code is calling "localization.set${attribute}".

Should we update the method for this call?

kdivya153: Resolved

defaultModule/requestModules/fcsSetters_firebolt.js
kdivya153: Addressed

defaultModule/requestModules/fcsSetters_firebolt.js
kdivya153: Addressed

kschrief (approved):

PR #324, opened 2025-02-03
fix: fcsSetter logic change from env to stack

kdivya153:


cypress/support/step_definitions/dynamicCalls.js
```diff
@@ -97,14 +97,12 @@ Given(
...
-          setParams = fireboltCallObject.setParams;
-
-          // Iterating through the object and invoke if it is a function
-          for (const key in setParams) {
-            if (typeof setParams[key] === CONSTANTS.TYPE_FUNCTION) {
-              setParams[key] = setParams[key]();
-            }
-          }
+          // Deep clone setParams while resolving functions to their return values
+          setParams = JSON.parse(
```
kdivya153, line 101: There was issue that If setParams in fireboltCall is an object, it was not re-evaluated in the next test case because it held a reference to functions. The changes ensure functions are executed, and setParams holds the latest resolved values instead of stale references.

cypress/support/cypress-support/src/globalSetter.js
```diff
@@ -1,19 +1,20 @@
...
-  await validateResponse(methodName, params, response);
-  console.log(`[${methodName}] ${message}`);
+  await validateResponse(params, response);
+  cy.log(`${message}`);
   return response;
 };
 
 global.setterFailure = (message, error) => {
-  const methodName = UTILS.getEnvVariable(CONSTANTS.REQUEST_OVERRIDE_METHOD);
+  fcsSetterStack.popMethod();
```
kschrief, line 16: Thinking more about this, wouldn't we not want to pop here but in cy.sendMessageToPlatforms()?

The actual logic processing the request shouldn't be responsible for the stack. SMTP should be responsible and the logic inside (Request overrides, FCS setters, etc) should just be referencing the stack.

cypress/support/cypress-support/src/main.js
```diff
@@ -254,11 +255,11 @@ export default function (module) {
     cy.wrap(requestMap, { timeout: CONSTANTS.SEVEN_SECONDS_TIMEOUT }).then(async (requestMap) => {
       return new Promise(async (resolve, reject) => {
         const [moduleName, methodName] = requestMap.method.split('.');
-        Cypress.env(CONSTANTS.REQUEST_OVERRIDE_METHOD, requestMap.method);
+        // Push method onto the stack to keep track of the current method
+        fcsSetterStack.pushMethod(requestMap.method);
```
kschrief, line 259: Linking to my comment above, we're pushing here before we call overrides, get responses, etc. We should be popping in this file after that logic, not during the FCS Setters logic itself.

cypress/support/step_definitions/dynamicCalls.js
kschrief: Has this been thoroughly tested? I don't think we should be taking this up in this PR.

kschrief (changes requested):

cypress/support/step_definitions/dynamicCalls.js
kdivya153: Yes Kurt,Vitalli also tested from his end and gave green

cypress/support/cypress-support/src/main.js
kdivya153: Intially,I also had the same thoughts but since we need both the methodNames need  inorder to validate response,it is difficult to pop in the SMTP.https://github.com/rdkcentral/firebolt-certification-suite/blob/16914ce7d61eee066ce867bcdd718a5e0bb34bbb/cypress/support/cypress-support/src/globalSetter.js#L55

cypress/support/cypress-support/src/main.js
```diff
@@ -299,6 +299,7 @@ export default function (module) {
           } else {
             resolve(null);
           }
+          fcsSetterStack.popMethod();
```
kschrief, line 302: Looks like we're only "popping" calls that aren't FCS Setters but we're pushing all methods.

We need to pop all calls that go in, not just the non-FCS ones.

cypress/support/validations/schemaValidation.js
```diff
@@ -176,6 +176,9 @@ Cypress.Commands.add('getSchema', (methodOrEvent, params, schemaType) => {
       cy.fixture(CONSTANTS.OPENRPC_ERROR_SCHEMA_PATH).then((errorSchemaObject) => {
         return errorSchemaObject.errorSchema;
       });
+    } else if (methodOrEvent.startsWith(CONSTANTS.FCS_SETTER)) {
```
kschrief, line 179: Can we just put a little comment here saying why we're putting this logic in?

kschrief (changes requested):

cypress/support/validations/schemaValidation.js
kdivya153: Addressed

cypress/support/cypress-support/src/main.js
kdivya153: Since fcsSetter makes a recursive call to sendMessagetoPlatforms (SMTP), if popMethod() is placed at the end of SMTP, it gets executed before the recursive SMTP call fully completes. This causes premature stack cleanup. To fix this, I tweaked the code to wait until the returned promise resolves (ensuring the recursive call finishes), and only then clean up the stack before exiting SMTP.

kschrief (approved):

PR #329, opened 2025-02-10
fix: Moved interaction service changes to config helper

Abhishk123:


kschrief (commented):
Looks good but I'm holding off final approval until the design question of allowing users to load their own cypress commands is addressed.

In addition do we have anywhere where we document what FCS does when it starts up? Documenting that a user can override cy.startAdditionalServices() is a good idea.

cypress/support/cypress-support/src/main.js
```diff
@@ -471,6 +438,27 @@ export default function (module) {
     return message;
   });
 
+  /**
+   * @module main
+   * @function runIntentAddon
+   * @description  If there was an add-on function in our config module it will be invoked and added additional fields else it will return the passed intent back.
+   * @param {string} input - Add-on function name present in config module.
```
neeradanelxsi, line 445: Line 444 is wrong 

did not understand the purpose of the env Cypress.env('startAdditionalServices')

Abhishk123: Updated the comments and added an environment check to allow the execution of other functions if needed

kschrief (dismissed):

ibrahimshuail (approved):

PR #330, opened 2025-02-11
feat: Firebolt v2 event handling

Eswar2103:


cypress/support/cypress-support/src/main.js
```diff
@@ -105,6 +105,21 @@ export default function (module) {
     const flattedOpenRpc = UTILS.getEnvVariable(CONSTANTS.DEREFERENCE_OPENRPC);
     const unflattedOpenRpc = flatted.parse(flattedOpenRpc);
     Cypress.env(CONSTANTS.DEREFERENCE_OPENRPC, unflattedOpenRpc);
+
+    // Check if the incoming SDK version is 2.0.0 or above and replace the widget in device with 2.0 changes.
```
jbigel, line 109: Do we plan to use the commented out code in the future?

cypress/support/cypress-commands/commands.js
```diff
@@ -1555,6 +1554,10 @@ Cypress.Commands.add('methodOrEventResponseValidation', (validationType, request
             ) {
               response.result = response.result.eventResponse;
             }
+            // Parse v2 events for first party events
```
kschrief, line 1557: What's the difference between 1st and 3rd party events?

cypress/support/cypress-support/src/main.js
```diff
@@ -200,6 +215,36 @@ export default function (module) {
     });
   });
 
+  afterEach(() => {
+    // Make a clear all event listeners call and clear the deregister the events
+    // Need to see what method name can be passed here, instead of device.name.
+    const requestMap = {
+      method: 'call.clearAllListeners',
```
kschrief, line 222: Is this call supported in both BOLT and FCA?

jbigel (approved):

cypress/support/appObjectConfigs.js
```diff
@@ -51,10 +50,10 @@ class eventObject {
   // Function to update the event response in event object.
   setEventResponseData(response, isNullCase, eventName) {
     if (
-      (response && response.eventResponse && response.eventResponse[this.eventObjectId] != null) ||
+      (response && response.eventResponse && response.eventResponse[this.eventName] != null) ||
```
kevinshahfws, line 53: I see this has been updated to eventName instead of eventObjectId... Won't this only be for v2.x? what about 1.x support?

kevinshahfws (approved):

cypress/support/cypress-commands/commands.js
```diff
@@ -1568,6 +1567,12 @@ Cypress.Commands.add('methodOrEventResponseValidation', (validationType, request
...
+            if (appId === UTILS.getEnvVariable(CONSTANTS.FIRST_PARTY_APPID) && UTILS.getEnvVariable(CONSTANTS.IS_BIDIRECTIONAL_SDK)) {
+              // "result": {"key": "new value"}
+              // parse the v2 event response from {"key": "new value"} to "new value"
+              response.result = Object.values(response.result)[0];
```
kschrief, line 1574: If we're expecting the request as {"key": "new value"} we should be parsing it properly. If for some reason this payload is passed with multiple keys we risk grabbing the wrong thing.

kschrief (changes requested):

cypress/support/appObjectConfigs.js
Eswar2103: Both 1.x and 2.x not supports with event name, we are clearing event listeners and map in after each, so event name will support both

kschrief (approved):

PR #333, opened 2025-02-12
feat: Added logic to override the fcsSetter methods

kdivya153:


cypress/support/cypress-support/src/utils.js
```diff
@@ -1217,6 +1217,39 @@ function censorPubSubToken(data) {
...
+    ? fireboltCallObject.overrides
+    : [fireboltCallObject.overrides];
+
+  for (const override of overrides) {
+    if (typeof override.applyWhen !== 'function') {
+      console.log("Ignoring override: Missing 'applyWhen()' function", override);
+      continue;
+    }
+
+    if (!override.applyWhen()) {
```
kschrief, line 1243: Let's surround this in a try/catch block. We have no control over what a user puts here so it's possible this will bubble up an error and we should handle it gracefully.

cypress/support/step_definitions/testSetup.js
```diff
@@ -48,14 +48,17 @@ Given(
...
-      // Retrieve the firebolt object from environment variables using the fireboltCallKey
-      const fireboltObject = UTILS.getEnvVariable(CONSTANTS.COMBINEDFIREBOLTCALLS)[fireboltCallKey];
-      if (fireboltObject) {
-        // Update the runtime environment variable with the firebolt object
-        runtime.fireboltCall = fireboltObject;
-        Cypress.env(CONSTANTS.RUNTIME, runtime);
-        fireLog.info(`Firebolt object successfully updated in runtime environment variable`);
-      }
+      // Retrieve the firebolt object from the fireboltCalls fixture
+      cy.getFireboltData(fireboltCallKey, CONSTANTS.SUPPORTED_CALLTYPES.FIREBOLTCALLS, false).then(
```
kschrief, line 52: Why was this changed?

kschrief (changes requested):

cypress/support/step_definitions/testSetup.js
kdivya153: Previously, the logic retrieved the fireboltObject from the env variable. Since we now have getFireboltData() to fetch the fireboltObject, I updated the logic to use getFireboltData() for consistency.
Additionally, if these changes were not made, applyOverrides() would be called in multiple places. With this change, applyOverrides() is now called only inside getFireboltData().

cypress/support/cypress-support/src/utils.js
kdivya153: Agreed,Wrapped with try/catch block.

cypress/support/step_definitions/testSetup.js
kschrief: Ahh, yes. I forgot this conversation. Thanks!

PR #349, opened 2025-02-26
feat: filtering validation logic

kdivya153:


cypress/support/cypress-commands/commands.js
```diff
@@ -1788,3 +1794,41 @@ Cypress.Commands.add('extractAppMetadata', (appDataDir, appMetaDataFile) => {
...
+ * - If 'includeValidations[key]' exists and does not contain 'value', validation is skipped ('false').
+ * - Otherwise, validation is performed ('true').
+ * @param {string} key - The key representing the type of validation.
+ * @param {any} value - The value to validate.
+ * @returns {boolean} 'true' if validation should proceed, 'false' for validation to skip.
+ */
+
+const shouldPerformValidation = (key, value) => {
+  // excludeValidations or includeValidations is not defined in the environment, assign {} to continue normal validations.
+  const excludeValidations = UTILS.getEnvVariable(CONSTANTS.EXCLUDE_VALIDATIONS, false) || {};
```
kschrief, line 1814: Let's just check to make sure these are valid JSON objects and attempt to "JSON.parse()" them if they're strings. I think if someone passes the JSON on the commandline it will come through as a string that needs to be converted to an object.

kschrief (changes requested):

cypress/support/cypress-commands/commands.js
kdivya153: Addressed

kschrief (approved):

PR #360, opened 2025-03-11
feat: Handle capture screenshot and validate

jbigel:


cypress/support/cypress-support/src/utils.js
```diff
@@ -1257,6 +1258,36 @@ function applyOverrides(fireboltCallObject) {
   return fireboltCallObject; // Return the original or modified object based on the override
 }
```
neeradanelxsi, line 1259: this function needs function description

cypress/support/cypress-support/src/utils.js
```diff
@@ -1257,6 +1258,36 @@ function applyOverrides(fireboltCallObject) {
...
```
neeradanelxsi, line 1260: i believe in validations.md file there are details about screenshot validations. Can we update the readme too

neeradanelxsi (dismissed):

kschrief (approved):

PR #369, opened 2025-03-18
fix: Adding unsub operation to transport layer for improved reliability

parag-pv:
Unsubscribe operation as a supporting change to queue implementation for platform communication.


cypress/support/cypress-support/src/transport.js
```diff
@@ -65,6 +65,19 @@ export default class Transport {
     }
   }
 
+  async unsubscribe() {
+    try {
+      if (process.env.transportObject) {
+        const transportObjectArray = process.env.transportObject;
```
kschrief, line 71: Are we sure this is an array? We should check and convert if necessary.

kschrief (changes requested):

kschrief (approved):

PR #375, opened 2025-03-21
fix: To support external sdk

kdivya153:


cypress/support/cypress-commands/commands.js
```diff
@@ -68,73 +68,76 @@ Cypress.Commands.add(
...
-          params = UTILS.getEnvVariable(envParam, false);
-        }
-        // If params contain CYPRESSENV in any parameter assigning corresponding env value
-        if (Array.isArray(params)) {
-          params.forEach((item) => {
-            const containEnv = Object.keys(item).find((key) => key.includes('CYPRESSENV'));
-
+Cypress.Commands.add(
+  'fireboltDataParser',
+  (key, sdk = UTILS.getEnvVariable(CONSTANTS.SUPPORTED_SDK)[0]) => {
```
ksentak, line 73: For the fireboltDataParser command I would assume that the default SDK thats passed in would be `Firebolt`. If that's the case why do we need to call `UTILS.getEnvVariable(CONSTANTS.SUPPORTED_SDK)[0]`. What happens if someone changes the order of the array for CONSTANTS.SUPPORTED_SDK? I know this wasn't code you originally added but it's something we should consider.

cypress/support/cypress-commands/commands.js
```diff
@@ -68,73 +68,76 @@ Cypress.Commands.add(
...
+    const results = [];
+    Cypress.env(CONSTANTS.IS_SCENARIO_EXEMPTED, false);
+    if (UTILS.getEnvVariable(CONSTANTS.SUPPORTED_SDK).includes(sdk)) {
```
ksentak, line 76: We ran into issues with .includes() the tother day with MF. To be super safe, can we make check this on two lines?

```javascript
const supportedSDK = UTILS.getEnvVariable(CONSTANTS.SUPPORTED_SDK);
if (Array.isArray(supportedSDK) && supportedSDK.includes(sdk)) {

```

cypress/support/step_definitions/validations.js
```diff
@@ -42,7 +42,7 @@ import UTILS, { fireLog } from '../cypress-support/src/utils';
 Given(
   /'(.+)' platform (responds|triggers|does not trigger)(?: to '(.+)')? (with|for|event)(?: for)? '(.+)'$/,
   async (sdk, eventExpected, appId, event, key) => {
-    if (CONSTANTS.SUPPORTED_SDK.includes(sdk)) {
+    if (UTILS.getEnvVariable(CONSTANTS.SUPPORTED_SDK).includes(sdk)) {
```
ksentak, line 45: Same comment as above

```javascript
const supportedSDK = UTILS.getEnvVariable(CONSTANTS.SUPPORTED_SDK);
if (Array.isArray(supportedSDK) && supportedSDK.includes(sdk)) {

```

ksentak (changes requested):
Small changes requested.

cypress/support/cypress-commands/commands.js
kdivya153: For the fireboltDataParser command I would assume that the default SDK thats passed in would be Firebolt. If that's the case why do we need to call UTILS.getEnvVariable(CONSTANTS.SUPPORTED_SDK)[0]. What happens if someone changes the order of the array for CONSTANTS.SUPPORTED_SDK? I know this wasn't code you originally added but it's something we should consider.

cypress/support/step_definitions/validations.js
kdivya153: Addressed

ksentak (approved):

PR #377, opened 2025-03-25
fix: Screenshot link provided in report opens to 'Resource not found'

PreethiMaai:


neeradanelxsi (approved):

jbigel (dismissed):

reportTemplates/generic.js
```diff
@@ -102,7 +102,11 @@ attachScreenshotLinks = () => {
     if (urlPattern.test(preHTML)) {
       // Replace plain text URLs with clickable hyperlinks
       const updatedHTML = preHTML.replace(urlPattern, (url) => {
-        return '<a href="' + url + '" target="_blank" rel="noopener noreferrer">' + url + '</a>';
+        let cleanedUrl = url;
```
ksentak, line 105: I recommend a different approach. I suggest modifying the regex on line 61 to be:

```javascript
const urlPattern = /(https:\/\/[^\s]+\.jpg)/g; 
```
This ensures that only URLs ending with .jpg are matched, making the pattern more precise and aligned with the expected use case. If there is text added before the `]` but after the URL, the current split(']')[0] approach will not work.

ksentak (changes requested):

reportTemplates/generic.js
PreethiMaai: code has been updated as suggested and ensured it's working

ksentak:
Approved!

PR #383, opened 2025-04-01
feat: Improve logging

ksentak:


cypress/support/cypress-support/src/main.js
```diff
@@ -421,6 +429,9 @@ export default function (module) {
 
       if (!UTILS.getEnvVariable(CONSTANTS.DEVICE_MAC)) {
         cy.log(CONSTANTS.DEVICE_MAC_UNAVAILABLE).then(() => {
+          logger.error(
```
kschrief, line 432: Let's set this as the assert failure message so the user will more easily see it.

kschrief (changes requested):

kschrief (dismissed):

PR #393, opened 2025-04-10
feat: To update runInfo for AppAssurance

kdivya153:


cypress/support/cypress-support/src/main.js
```diff
@@ -566,6 +566,33 @@ export default function (module) {
...
+   * @module commands
+   * @function callConfigModule
+   * @description Check the configModule for the function and call it with the params.
+   * @param {String} methodName - Name of the function to be called from the config module.
+   * @param {...*} params - The parameters required to perform the function.
+   * @example
+   * cy.callConfigModule('getReportData', param1, param2)
+   */
+
+  Cypress.Commands.add('callConfigModule', (methodName, ...params) => {
```
kschrief, line 579: Can we add a "module" param here as well with a default "additionalServices"? We're planning on expanding this beyond just additionalServices

cypress/support/cypress-support/src/main.js
```diff
@@ -566,6 +566,33 @@ export default function (module) {
...
+    console.log(`Divya: Calling "${methodName}" from configModule...`);
+
+    return cy.then(() => {
+      // Check if additionalServices exists inside configModule and function exists
+      const configFunction = module?.additionalServices?.[methodName];
+
+      if (typeof configFunction !== 'function') {
+        console.log(`additionalServices or function not found in the configModule`);
```
kschrief, line 587: Can we call out the module/function that wasn't found?

kschrief (changes requested):

cypress/support/cypress-support/src/main.js
kdivya153: To include moduleName, Since we’re using rest parameters (i.e., ...args), it becomes ambiguous to distinguish between parameters  a  moduleName.Should I switch to a named parameters object. Something like below while calling the function.
cy.callConfigModule({
  methodName: 'getReportData',
  params: ['abc', 123],                  //optional
  moduleName: 'additionServices'  //optional 
});

PR #396, opened 2025-04-14
feat: Auto-detect action

AdityaKasar:
Auto-detect action based on feature file.

### DIFF
--- a/cypress/support/cypress-support/src/utils.js
+++ b/cypress/support/cypress-support/src/utils.js
@@ -1240,6 +1240,31 @@ function captureScreenshot() {
   }
 }
 
+/**
+ * Determines the SDK action type based on the feature file name
+ * from the current test's title path.
+ *
+ * @param {Object} testRunnable - The current test context from Cypress (e.g., cy.state("runnable")).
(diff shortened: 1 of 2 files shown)


cypress/support/cypress-support/src/utils.js
```diff
@@ -1240,6 +1240,34 @@ function captureScreenshot() {
...
+  } else if (/Firebolt Certification Manage-SDK validation/i.test(featureFile)) {
+    console.log("Matched Manage-SDK suite. Action: 'MANAGE'");
+    return "MANAGE";
+  } else if (/Firebolt Certification Discovery-SDK validation/i.test(featureFile)) {
+    console.log("Matched Discovery-SDK suite. Action: 'DISCOVERY'");
+    return "DISCOVERY";
+  }
+
+  console.log("No matching SDK suite found. Action: null");
+  return null;
```
kdivya153, line 1266: Logic looks good to me,Can you double check with Kurt whether it has to return null?Probably we have to return default action(core)/NA

AdityaKasar: I agree, I will change it to default core

kschrief (dismissed):

PR #401, opened 2025-04-17
feat: Continue watching implementation

Abhishk123:


cypress/support/step_definitions/validations.js
```diff
@@ -336,3 +336,34 @@ Given(
...
+/**
+ * @module ValidationGlue
+ * @function I verify '(.+)' rail is(?: (.+))? visible '(.+)'
+ * @description 
+ * @param {String} testType - 
+ * @param {String} notVisible -
+ * @param {String} input -
+ * @example
+ * Then '3rd party app' will stay in 'foreground' state
+ * Then '3rd party app' will be in 'background' state
```
neeradanelxsi, line 349: 348, 349 need to be corrected

Nandana-NNR: addressed

cypress/support/constants/constants.js
```diff
@@ -385,6 +385,7 @@ module.exports = {
     GETAPPDATA: 'fcs.getAppData',
     SENDKEYPRESS: 'fcs.sendKeyPress',
     SENDVOICECOMMAND: 'fcs.sendVoiceCommand',
+    FINDLOGPATTERN: 'fcs.findLogPattern',
```
jbigel, line 388: Again this seems like a good candidate for `cy.callConfigModule`

cypress/support/step_definitions/validations.js
```diff
@@ -336,26 +336,3 @@ Given(
     });
   }
 );
-
```
jbigel, line 339: Why are we removing this?

jbigel (changes requested):

cypress/support/step_definitions/validations.js
jbigel: Why do we need to move this to configHelper? There is nothing proprietary in it.

PR #405, opened 2025-04-25
feat: Added configuarable time out

Abhishk123:


cypress/support/cypress-support/src/main.js
```diff
@@ -279,7 +285,7 @@ export default function (module) {
               // Perform MTC/FB call only if the message is not null
               if (message != null) {
                 return transport
-                  .sendMessage(message)
+                  .sendMessage(message, responseWaitTime)
```
jbigel, line 288: Should this be `validResponseWaitTime`?

cypress/support/cypress-support/src/main.js
```diff
@@ -415,11 +423,14 @@ export default function (module) {
         });
       }
 
-      cy.sendMessagetoApp(requestTopic, responseTopic, intent).then((response) => {
+      cy.sendMessagetoApp(requestTopic, responseTopic, intent, sanityTimeout).then((response) => {
         cy.log('Response from Firebolt Implementation: ' + response);
 
         if (response === CONSTANTS.RESPONSE_NOT_FOUND) {
-          assert(false, CONSTANTS.NO_MATCHED_RESPONSE);
+          assert(
```
jbigel, line 430: Could we use fireLog.fail here?

cypress/support/cypress-support/src/main.js
```diff
@@ -500,53 +510,54 @@ export default function (module) {
...
+              // Response recieved from queue
+              logger.debug(`Response received from queue: ${JSON.stringify(results)}`);
+              return results;
+            } else if (Cypress.env(CONSTANTS.IS_RPC_ONLY)) {
+              return true;
+            }
+          });
+      } else {
+        cy.log(CONSTANTS.APP_TRANSPORT_UNAVAILABLE).then(() => {
+          assert(false, CONSTANTS.APP_TRANSPORT_UNAVAILABLE);
```
jbigel, line 556: Could we use fireLog.fail here?

jbigel (changes requested):

cypress/support/cypress-support/src/main.js
Abhishk123: No we need to pass `responseWaitTime` here. Changes the var name from `validResponseWaitTime` to `requestTimeout` to avoid confusion

cypress/support/cypress-support/src/main.js
Abhishk123: addressed the comment

cypress/support/cypress-support/src/main.js
jbigel: I'm just trying to understand why we do parsing logic on lines 243-246 but here we pass it blindly? Is that because it can not be passed and we need to assign a timeout for the the wrap but we can pass a null param to sendMessage?

It can be a little confusing reading the code so can we add a comment explaining why we parse it but blindly pass it here?

Abhishk123: 1. Parsing `requestTimeout` was added to prevent getting `NaN` when no wait time is provided for the `sendMessagetoPlatforms` command. This is specifically used for the wrap command.
2. `responseWaitTime` is passed directly to `sendMessage` because the default time is managed by the transport layer.

cypress/support/cypress-support/src/main.js
```diff
@@ -232,13 +232,21 @@ export default function (module) {
...
    * @example
    * cy.sendMessagetoPlatforms({"method": "closedCaptioning", "param": {}})
+   * cy.sendMessagetoPlatforms({"method": "closedCaptioning", "param": {}}, 20000)
    */
 
-  Cypress.Commands.add('sendMessagetoPlatforms', (requestMap) => {
-    return cy.wrap(requestMap, { timeout: 75000 }).then({ timeout: 75000 }, () => {
+  Cypress.Commands.add('sendMessagetoPlatforms', (requestMap, responseWaitTime) => {
+    // The requestTimeout is calculated to ensure a valid timeout value is passed to the wrap function.
+    // However, the responseWaitTime is undefined, it is handled by the transport layer.
```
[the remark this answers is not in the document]
Abhishk123: updated the comment

jbigel (approved):

ibrahimshuail (approved):

PR #415, opened 2025-05-07
feat: Propogate cypress exit code

AdityaKasar:


AdityaKasar:
Any specific reason for not adding `process.exit(1);` in this block for both run() and open()

`  
    cypressProcess.on('error', (error) => {
    console.error(`Error: ${error.message}`);
  });
`

PR #427, opened 2025-05-20
feat: Singleton PubSub Client

AdityaKasar:


cypress/support/cypress-support/src/main.js
```diff
@@ -37,13 +37,14 @@ const internalV2FireboltMockData = require('../../../fixtures/fireboltCalls/inde
...
 export default function (module) {
+  let pubSubClient = null;
   const config = new Config(module);
   const validationModule = new Validation(module);
   const transport = new TransportLayer();
   Cypress.env(CONSTANTS.RESPONSE_TOPIC_LIST, []);
 
   // Fetch the required appTransport from config module
-  appTransport = module.appTransport;
+  appTransport = module.externalTransport.Linchpin.default;
```
ksentak, line 47: This cant be used here. Need to remove.

cypress/support/cypress-support/src/main.js
```diff
@@ -532,13 +542,13 @@ export default function (module) {
...
-        appTransport.subscribe(responseTopic, UTILS.subscribeResults);
+        pubSubClient.subscribe(responseTopic, UTILS.subscribeResults);
         UTILS.getEnvVariable(CONSTANTS.RESPONSE_TOPIC_LIST).push(responseTopic);
       }
 
-      if (appTransport) {
+      if (pubSubClient) {
         // Publish the message on topic
-        appTransport.publish(requestTopic, JSON.stringify(intent), headers);
+        pubSubClient.publish(requestTopic, JSON.stringify(intent), headers);
```
ksentak, line 551: Let's clean this up. We check !pubSubClient then later in the same function we check if pubSubClient. Rework so there is only 1 if check.

cypress/support/cypress-support/src/main.js
```diff
@@ -554,7 +564,7 @@ export default function (module) {
           });
       } else {
         cy.log(CONSTANTS.APP_TRANSPORT_UNAVAILABLE).then(() => {
-          fireLog.fail(CONSTANTS.APP_TRANSPORT_UNAVAILABLE);
+          assert(false, CONSTANTS.APP_TRANSPORT_UNAVAILABLE);
```
ksentak, line 567: Why did we change from fireLog.fail to assert here?

ksentak (changes requested):
Left a couple of comments for requested changes. Also, I want to have conversations about why there needed to be changes to how parsing was done for healthcheck and report. This alerts me that something might not be right. Perhaps what we incorrectly changed how an incoming message is parsed or something so we need to make sure that isn't the case.

cypress/support/cypress-support/src/main.js
AdityaKasar: Yes, this one got missed. I will fix it.

AdityaKasar:
> Also, I want to have conversations about why there needed to be changes to how parsing was done for healthcheck and report. This alerts me that something might not be right. Perhaps what we incorrectly changed how an incoming message is parsed or something so we need to make sure that isn't the case

Updated _handleIncomingMessage() logic in Websocket to handle response from (app) `comcast.test.firecer` differently.
`          if(data.topic.includes("comcast.test.firecert")){
            callback(typeof payload === 'string' ? payload : JSON.stringify(payload), data.headers);
          }else{
            callback(typeof data === 'string' ? data : JSON.stringify(data), data.headers);
          }`

cypress/support/cypress-support/src/main.js
```diff
@@ -37,29 +37,31 @@ const internalV2FireboltMockData = require('../../../fixtures/fireboltCalls/inde
...
 export default function (module) {
+  let pubSubClient = null;
   const config = new Config(module);
   const validationModule = new Validation(module);
   const transport = new TransportLayer();
   Cypress.env(CONSTANTS.RESPONSE_TOPIC_LIST, []);
 
   // Fetch the required appTransport from config module
-  appTransport = module.appTransport;
+  const firstExternalTransportKey = Object.keys(module.externalTransport)[0];
```
ksentak, line 47: This should work without issue. In the config helper we may want to make a comment to ensure the order of the exports stays the same. Otherwise this would break. Although I dont anticipate any other transportation methods.

cypress/support/cypress-support/src/main.js
```diff
@@ -532,31 +539,25 @@ export default function (module) {
...
+      return UTILS.getEnvVariable(CONSTANTS.MESSAGE_QUEUE)
+        .LongPollQueue(headers.id, longPollTimeout)
+        .then((results) => {
+          if (results) {
+            // Response recieved from queue
+            logger.debug(`Response received from queue: ${JSON.stringify(results)}`);
+            return results;
+          } else if (Cypress.env(CONSTANTS.IS_RPC_ONLY)) {
+            return true;
+          }
```
ksentak, line 559: We should still keep this else

```
else {
        cy.log(CONSTANTS.APP_TRANSPORT_UNAVAILABLE).then(() => {
          fireLog.fail(CONSTANTS.APP_TRANSPORT_UNAVAILABLE);
        });
      }
```

ksentak (changes requested):
Minimal changes requested here. Otherwise looks good!

cypress/support/cypress-support/src/main.js
kschrief: This is problematic. It works for **us** but any future config modules would be blind to this issue.

There needs to be a way.

Created a ticket to deal with this at a later time.

kschrief: This is problematic. It works for us but any future config modules would be blind to this issue.

This needs to be resolved before this is merged in.

kschrief (changes requested):

cypress/support/cypress-support/src/main.js
AdityaKasar: My understanding is that code will not reach this else condition as this block exists at the beginning of the function

`     if (!pubSubClient) {
        cy.log(CONSTANTS.APP_TRANSPORT_UNAVAILABLE).then(() => {
          fireLog.fail(CONSTANTS.APP_TRANSPORT_UNAVAILABLE);
        });
        return;
      }`

cypress/support/cypress-support/src/main.js
AdityaKasar: made changes to address the review comment.

cypress/support/cypress-support/src/main.js
AdityaKasar: changed to PubSub.default

ksentak (dismissed):

kschrief (changes requested):
I see no changes in the default config module. This must be taken into account before we can approve this.

PR #430, opened 2025-05-26
feat: content validation for events when triggered in interaction logs

anjalimukundan:


cypress/support/cypress-commands/commands.js
```diff
@@ -1043,7 +1043,7 @@ Cypress.Commands.add('launchApp', (appType, appCallSign, deviceIdentifier, inten
         `No intentTemplates found for ${appType}, make sure the intentTemplates are defined as per the appType`
       );
     }
-
+    Cypress.env(CONSTANTS.RUNTIME).intentTemplate = intentTemplate;
```
Eswar2103, line 1046: Just wanted to know why this is added here, means was it a miss previously to add intent templates in runtime or for current implementation it is required?

PR #436, opened 2025-06-05
feat: Added getSchemaFromOpenRpc command

Abhishk123:


cypress/support/validations/schemaValidation.js
```diff
@@ -212,6 +212,48 @@ Cypress.Commands.add('getSchema', (methodOrEvent, params, schemaType) => {
   });
 });
 
+/**
+ * @module schemaValidation
+ * @function getSchemaFromOpenRpc
+ * @description get the schema from OpenRPC based on method name and type.
+ * @param {string} methodName - The name of the method to search in OpenRPC.
+ * @param {string} type - The type of schema to return, either "result" or "listen".
```
kschrief, line 220: This only applies to events. The doc should mention that.

cypress/support/validations/schemaValidation.js
```diff
@@ -212,6 +212,48 @@ Cypress.Commands.add('getSchema', (methodOrEvent, params, schemaType) => {
...
+
+    // Check if method is an event (has a tag with name 'event')
+    const isEvent = Array.isArray(method.tags) && method.tags.some((tag) => tag.name === 'event');
+
+    if (isEvent) {
+      if (type === CONSTANTS.RESULT) {
+        // For events, find x-subscriber-for and return referenced method's result.schema
+        const eventTag = method.tags.find((tag) => tag.name === 'event' && tag['x-subscriber-for']);
+        if (eventTag) {
+          const subscribeForName = eventTag['x-subscriber-for'];
```
kschrief, line 239: We already know that there are currently methods out there that don't follow this tagging. We should properly log when an event doesn't contain an "x-subscriber-for" tag before we return null.

kschrief (changes requested):

cypress/support/validations/schemaValidation.js
Abhishk123: we can use it for methods also

neeradanelxsi (approved):

jbigel (approved):

kevinshahfws (approved):

PR #441, opened 2025-06-19
feat: Setup: Lifecycle base/child classes and object creation

Abhishk123:


cypress/support/cypress-support-lifecycle/1.x/lifecycleConfig.json
```diff
@@ -0,0 +1,54 @@
+{
+  "validationConfig": {
+    "initializing": {
+      "stateValidation": { "req": "R*2.1.1 & R*2.1.2 & R*4.6" },
+      "historyValidation": { "req": "R*2.1.3" },
```
neeradanelxsi, line 5: i believe the plan was to keep a lifecycle folder under cypress-support instead of creating a new folder cypress-support-lifecycle

cypress/support/cypress-support-lifecycle/1.x/lifecycleConfig.json
```diff
@@ -0,0 +1,54 @@
+{
+  "validationConfig": {
+    "initializing": {
```
neeradanelxsi, line 3: I believe validationConfig is not required. If not, please remove it after confirmation

cypress/support/cypress-support-lifecycle/1.x/lifecycleConfig.json
Abhishk123: kurt suggested both places, created a lifecycle folder inside `cypress-support/src`, this looks better

cypress/support/cypress-support-lifecycle/1.x/lifecycleConfig.json
Abhishk123: I don't see any usage, so removed it

cypress/support/cypress-support/src/lifecycle/1.x/lifecycle_v1.js
```diff
@@ -0,0 +1,90 @@
+const lifecycleConfig = require('./lifecycleConfig.json');
+const { LifeCycleAppConfigBase } = require('../LifeCycleAppConfigBase');
+const logger = require('../../../../Logger')('lifecycle_v1.js');
+const CONSTANTS = require('../../../../constants/constants');
+
+class notificationConfig {
```
kschrief, line 6: This is identical in V1 and V2. Can these classes be a part of the base class?

cypress/support/cypress-support/src/lifecycle/1.x/lifecycle_v1.js
```diff
@@ -0,0 +1,90 @@
...
+  constructor(message) {
+    this.time = Date.now();
+    this.message = message;
+  }
+}
+
+class stateConfig {
```
kschrief, line 13: This is identical in V1 and V2. Can these classes be a part of the base class?

cypress/support/cypress-support/src/lifecycle/1.x/lifecycle_v1.js
```diff
@@ -0,0 +1,90 @@
...
+  setAppState() {
+    // TODO: Implement V1 specific flow for setting app state
+  }
+
+  // Function to set a new state for the appObject following below rules:
+  // Rule 1: Only allow a state transition if the new requested state is not the same as current state.
+  // Rule 2: If the setAppObjectState is for initializing state and if history list is empty, initializing state will not be pushed to history
+  // Rule 3: Check the look up table in appObjectConfigData for allowed state transitions before pushing to history
+  // Rule 4: If setAppObjectState is called after initializing, push initializing state to history before the new requested state
+  setAppObjectState(newState) {
```
kschrief, line 48: This function is identical in V1 and V2. Can this be put in the base class?

kschrief (changes requested):

cypress/support/cypress-support/src/lifecycle/1.x/lifecycle_v1.js
Abhishk123: In v2 new list we are going to add as part of setAppState, so i kept in both files

cypress/support/cypress-support/src/lifecycle/1.x/lifecycle_v1.js
Abhishk123: In v2, we might need to update this. Once setAppState changes are added, and if both files contain the same code, we will move it to a base class file.

cypress/support/cypress-support/src/lifecycle/1.x/lifecycle_v1.js
Abhishk123: In v2, we might need to update this. Once setAppState changes are added, and if both files contain the same code, we will move it to a base class file.

ibrahimshuail (approved):

neeradanelxsi (approved):

ibrahimshuail (approved):

PR #442, opened 2025-06-20
feat: Dynamic intent fetching from XVP asset generator

AdityaKasar:


cypress/support/cypress-support/src/main.js
```diff
@@ -606,15 +606,13 @@ export default function (module) {
   Cypress.Commands.add(
     'callConfigModule',
     (methodName, params = [], moduleName = 'additionalServices') => {
-      console.log(`Calling "${methodName}" from configModule.${moduleName} with params:`, params);
-
-      return cy.then(() => {
+      return cy.then(async () => {
```
jbigel, line 609: Was there a reason for removing this log?

cypress/support/cypress-support/src/main.js
```diff
@@ -606,15 +606,13 @@ export default function (module) {
...
         const configFunction = module?.[moduleName]?.[methodName];
         if (typeof configFunction !== 'function') {
           console.log(`${moduleName}.${methodName} not found in the config module.`);
           return null;
         }
-        return configFunction(...params);
+        return await configFunction(...params);
```
jbigel, line 615: Since this is an existing function it would be better to handle any waiting for returns by the caller instead inside the command directly to ensure we do not impact any existing code.

cypress/support/cypress-support/src/utils.js
```diff
@@ -1288,6 +1288,33 @@ function determineActionFromFeatureFile(testRunnable) {
...
 }
 
+/**
+ * Builds intent environment and resolves intent using Cypress config module.
+ * @param {string} appId - The durable app ID.
+ * @param {string} programType - The type of program (e.g., 'movie', 'episode').
+ * @returns {Chainable<any>} Cypress chainable for resolved intent.
+ */
+function buildAndResolveIntent(appId, programType) {
+  const runtimeEnv = Cypress.env();
```
jbigel, line 1298: We shouldn't need to do any of this. The env variables that are required should be able to be fetched in the configMod.

cypress/support/cypress-support/src/utils.js
```diff
@@ -1288,6 +1288,33 @@ function determineActionFromFeatureFile(testRunnable) {
...
+      durableAppId: appId,
+      vodShowings: 'all',
+      numShowings: 'all',
+    },
+    knownEntities: {},
+    app_metadata: {},
+    runtime: {},
+  };
+
+  return cy.callConfigModule('resolveIntent', [programType, intentEnv]);
```
jbigel, line 1315: We should be able to just add this call where buildAndResolveIntent is called. The building of the object is unnecessary. The only thing that needs to b e passed to the configMod is appId - everything else can be directly retrieved from cypress.env in configMod.

jbigel (changes requested):

cypress/support/cypress-support/src/main.js
AdityaKasar: removed by mistake. reverted

cypress/support/cypress-commands/commands.js
```diff
@@ -1044,6 +1044,18 @@ Cypress.Commands.add('launchApp', (appType, appCallSign, deviceIdentifier, inten
       );
     }
     Cypress.env(CONSTANTS.RUNTIME).intentTemplate = intentTemplate;
+
+    if (intentTemplate?.data?.programType) {
```
jbigel, line 1048: I'm not sure I understand this if condition.

PR #451, opened 2025-07-08
event validation changes

Nandana-NNR:


cypress/support/cypress-support/src/lifecycle/1.x/lifecycle_v1.js
```diff
@@ -228,6 +228,92 @@ class lifecycle_v1 extends LifeCycleAppConfigBase {
...
+                  fireLog.assert(false, pretext + ' : ' + CONSTANTS.FAIL);
+                }
+
+                pretext = id === undefined ? ' : Content ' : id + ' : Content ';
+                UTILS.assertWithRequirementLogs(
+                  pretext,
+                  JSON.stringify(newAppEvent.event),
+                  JSON.stringify(appObjectEvent.message)
+                );
+              }
```
neeradanelxsi, line 305: v1 and v2 events are different , in the sense 
onForeground event will be in the format {current : "foreground", previous "background"}
onActivate event wil have empty response 

So where are we doing conversion here ? in FCA or FCS / 
I believe this would affect both history and event validation

cypress/support/cypress-support/src/lifecycle/2.x/lifecycle_v2.js
```diff
@@ -187,4 +187,37 @@ export default class lifecycle_v2 extends LifeCycleAppConfigBase {
...
+        state: state
+      },
+    },
+    {
+      method: CONSTANTS.REQUEST_OVERRIDE_CALLS.THUNDEREVENTHANDLER,
+      params: {},
+      task: CONSTANTS.TASK.THUNDEREVENTHANDLER,
+    },
+  ];
+
```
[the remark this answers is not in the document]
Nandana-NNR: Unlike other validations, which make use of validationObjects and internal cy.customValidation function to perform validations, here in legacy lifecycle logic, we cannot follow that approach, that's why created request and response overrides since the code involves sending and receiving request to/from platform
But alternatively, I think we can make use of callConfigModule functionality to replace these as customValidations and call it using the callConfigModule functionality, so updated the code to implement the same. Note: We still need the request override to call sendMessagetoPlatform to publish message

cypress/support/cypress-support/src/lifecycle/1.x/lifecycle_v1.js
```diff
@@ -228,6 +228,92 @@ class lifecycle_v1 extends LifeCycleAppConfigBase {
...
+                let appObjectEvent;
+                if (eventIndex == 1) {
+                  appObjectEvent = appObject.state.notification[0];
+                } else {
+                  const appObjectStateItem =
+                    appObject.history[appObject.history.length - eventIndex];
+                  appObjectEvent = appObjectStateItem.notification[0];
+                }
+                // Perform schema and content validation of app event data against app object event data
+                const id = Array.isArray(eventId) ? eventId[eventIndex - 1] : eventId;
```
neeradanelxsi, line 290: @Nandana-NNR , can you just confirm that the existing logic for 1.x (present inside validateLifecycleHistoryAndEvents) is same here when you moved to this class?

Nandana-NNR: yes v1 logic is exactly the same..
only thing is in v1, validateLifecycleHistoryAndEvents was a single function. Now we have split into 2 seperate functions - validateHistory and validateEvents as per the the new requirement. validateHistory is covered in Abhishek's ticket - FIRECERT-3178

PR #456, opened 2025-07-15
Selective validation for thunder events

Nandana-NNR:


ibrahimshuail (approved):

cypress/support/cypress-support/src/lifecycle/2.x/lifecycle_v2.js
```diff
@@ -188,36 +189,112 @@ export default class lifecycle_v2 extends LifeCycleAppConfigBase {
...
-  const requestMaps = [
-    {
-      method: CONSTANTS.REQUEST_OVERRIDE_CALLS.VALIDATELIFECYCLEFIREBOLTLOGS,
-      params: {
-        appId: appId,
-        state: state
+  // Validate lifecycle firebolt and thunder events
+  validateEvents(isEventsExpected) {
+    Cypress.env(CONSTANTS.IS_EVENTS_EXPECTED, isEventsExpected);
+    cy.getFireboltData(CONSTANTS.LIFECYCLE_EVENT_VALIDATION).then((fireboltData) => {
```
neeradanelxsi, line 195: we should not be assigning random values to env variables and use it elsewhere . cant this be passed to some function ?

cypress/support/cypress-commands/commands.js
```diff
@@ -1441,9 +1441,11 @@ Cypress.Commands.add('methodOrEventResponseValidation', (validationType, request
...
     const scenario = object.type;
     const tags = object.tags;
+    const assertionDef = object.assertionDef;
     // To check whether the validation should be performed or not based on the include/exclude valiodation object
-    if (!shouldPerformValidation('validationTypes', scenario)) return;
-    if (!shouldPerformValidation('validationTags', tags)) return;
+    if (!UTILS.shouldPerformValidation('validationTypes', scenario)) return;
+    if (!UTILS.shouldPerformValidation('validationTags', tags)) return;
+    if (!shouldPerformValidation('customAssertionDefs', assertionDef)) return;
     if (scenario === CONSTANTS.SCHEMA_ONLY || !object.validations) return;
```
neeradanelxsi, line 1449: why is line 1448 not having UTILS ?

cypress/support/cypress-support/src/lifecycle/2.x/lifecycle_v2.js
```diff
@@ -188,36 +189,112 @@ export default class lifecycle_v2 extends LifeCycleAppConfigBase {
...
+                  if ('error' in parsed) {
+                    throw new Error(
+                      `Received error inside trigger event response for thunder state validation: ${JSON.stringify(parsed.error)}`
+                    );
+                  }
+
+                  return Array.isArray(parsed.result) ? parsed.result : [];
+                } catch (error) {
+                  throw new Error(
+                    `Received following error while parsing triggered event response  for thunder state validation - ${error.message}`
```
neeradanelxsi, line 266: can we discuss about the return format for the SMTP calls ?>

cypress/support/cypress-commands/commands.js
Nandana-NNR: corrected

cypress/support/cypress-support/src/lifecycle/2.x/lifecycle_v2.js
Nandana-NNR: sample response format shared for thunder event -
```
{
  "org.rdk.LifecycleManager.onAppLifecycleStateChanged": "{\"result\":[{\"eventResponse\":{\"method\":\"org.rdk.LifecycleManager.onAppLifecycleStateChanged\",\"params\":{\"appId\":\"\",\"appInstanceId\":\"\",\"oldLifecycleState\":\"\",\"newLifecycleState\":\"LOADING\"}}}]}"
}
```
To summarise,
Since the result is a string, we parse it and extract the eventResponses to an array along with error handling
We then reverse the eventResponse array so the latest occurence of event will be moved towards first and then filter the latest onAppLifecycleChanged event entry
Since the actual event response for an event is returned inside params field, we then parse it and extract state details and perform assertion

